PR #1, opened 2015-01-24
style class to crop with ellipsis

kellyredding:
This adds `.romo-crop-ellipsis`, which as it name implies, will crop
content just like `.romo-crop` but will also add ellipsis.  This
is done using `text-overflow: ellipsis`.

@jcredding ready for review.  You cool with this name (as opposed to `.romo-ellipsis` or something)?  I wanted to keep crop in there b/c this is about more than just adding the text overflow style.  You have to have the others with it or it won't work.  For reference:
- http://css-tricks.com/snippets/css/truncate-string-with-ellipsis/
- http://html5hub.com/ellipse-my-text/ (text-overflow: ellipsis section)

### DIFF
--- a/assets/css/romo/base.scss
+++ b/assets/css/romo/base.scss
@@ -415,8 +415,9 @@ h3 { @include text1; }
 
 .romo-match-line-height { line-height: 1 !important; }
 
-.romo-nowrap { white-space: nowrap !important; }
-.romo-crop   { white-space: nowrap !important; overflow: hidden !important; }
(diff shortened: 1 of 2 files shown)


jcredding:
@kellyredding - Looks good and I'm good with the `crop` in there. If you took out the "crop" overflow and whitespace CSS then I would think `romo-ellipsis` would make more sense but like you said that's not very useful. We'd always have to add `romo-crop romo-ellipsis`. Good stuff :boom:

PR #3, opened 2015-01-26
form js: add `submitCompleted` event

kellyredding:
This event will fire when the submit is finished, whether it was
successful or errored.  This is to make the API and feature set
more rich and so you can trigger behavior once the submit completes.

This also cleans up the indicator handling.  First it moves to triggering
the indicator before firing the before submit event.  This ensures
the very first and very last things done in the submit cylcle it
to start/stop the indicator.  This also updates the indicator logic
to run even if the element is disabled.  This is necessary as users
may wish to disable the indicator but still stop it later on.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/form.js
+++ b/assets/js/romo/form.js
@@ -68,8 +68,8 @@ RomoForm.prototype.onSubmitClick = function(e) {
 }
 
 RomoForm.prototype.doSubmit = function() {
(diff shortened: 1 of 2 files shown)


jcredding:
@kellyredding - Changes make sense and I like the addition of the complete event. Good stuff :boom:

PR #4, opened 2015-01-26
modal/dropdown js: clear content on close option

kellyredding:
This adds an option to make modals and dropdowns clear their contents
on close.  This is useful when many elems (maybe each on a row of
a list) all render the same markup.  A common example is each row of
a list having an edit modal form that contains identical markup
(submitting to unique urls).  If any elements of the form are referenced
with IDs they will be duplicated in the DOM once more than one modal
is invoked.  This allows you to clear the content on close to prevent
the duplication (since no two modals/dropdowns can be open at the
same time).

This also updates the modal/dropdown form components to default this
option to true if not already set.

@jcredding ready for review.


jcredding:

PR #5, opened 2015-01-26
dropdown js: set offset if setting position fixed

kellyredding:
This fixes a bug in Chrome where it doesn't properly honor the offset
it the dropdown element is dynamically changed to position: fixed.
This comes up when using dropdowns (for example select dropdowns)
in modals (ie model forms).  The initial opening of the dropdown
sets position fixed and calculates the offset.

It seems the calculating the offset the first time "resets" how
chrome determines the offset (ie honor postion fixed) b/c subsequent
offset calcs are done correctly.

This is somewhat a "hack" in that this behavior is not exhibited
in Firefox.  It is less of a hack though because we only apply it
when setting position fixed dropdowns (ie only when it is needed).

@jcredding ready for review.  Let me know if you have any better suggestion or if this is "too hacky" for you.

### DIFF
--- a/assets/js/romo/dropdown.js
+++ b/assets/js/romo/dropdown.js
@@ -260,6 +260,7 @@ RomoDropdown.prototype.onResizeWindow = function(e) {
 RomoDropdown.prototype.doPlacePopupElem = function() {
   if (this.elem.parents('.romo-modal-popup').size() !== 0) {
     this.popupElem.css({'position': 'fixed'});
+    this.popupElem.offset(this.elem.offset());
   }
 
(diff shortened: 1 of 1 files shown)


jcredding:
@kellyredding - You have "hock" in your commit message instead of "hack".

jcredding:
@kellyredding - This works for me :boom: I don't know of a better way to deal with this and it seems like its just a bug with Chrome. Based on your description it sounds like the kind of bug someone would accidentally put into a browser. I've seen other UI frameworks have these kind of HTML/CSS/JS hacks to  support different browsers.

PR #6, opened 2015-01-27
dropdown js: rework max height detection; make it work for position top

kellyredding:
This reworks the dropdown js to clean up how it detects the max
height available for the popup.  It also updates how this max height
applies to positioning the popup so that this calculation is honored
and works if positioning the popup on the top (previously positioning
top wouldn't benefit from max height detection).

This is prep for dropdowns switching their position if there isn't
enough height available for their popup.

@jcredding ready for review.


jcredding:
@kellyredding - Code looks good and its a nice feature to have for the edge cases that come up :boom:

The last line in your commit message says "switching their positions i there".

PR #7, opened 2015-01-27
dropdown js: position specific max available heights and padding

kellyredding:
This updates the max height detect logic to gather position
specific max available height and padding settings.  This is prep
for changing the position if there is not enough height to show
the dropdown.

@jcredding ready for review.


jcredding:

PR #8, opened 2015-01-27
dropdown js: switch position if it doesn't fit the configured way

kellyredding:
This updates the dropdown max height detect logic to automatically
switch the position if the dropdown won't fit in the configured
position.  If it fits in poth positions, it prefers the configured
way.  If it fits in neither position, it prefers the way with the
most available height.

@jcredding ready for review.


jcredding:
@kellyredding - Looks good :boom: I like the behavior, use what I say if you can, otherwise choose what will fit best.

PR #9, opened 2015-02-09
js: kepress component

kellyredding:
This component binds to an element and detects and publishes trigger
events on keypress events.  It is auto-invoked like other romo
components, publishes events and generally behaves like other romo
components.

This is a utility component.  It is designed to be used as a part
of bigger js components or behavior.  For example:
- submit-as-you-type form input
- pickers that search as you type
- keyboard shortcut handlers
- etc

@jcredding ready for review.  Please feedback on my naming choices and overall design.  Does it seem like this _could_ be integrated into the examples I mention above?  Thanks.

### DIFF
--- a/assets/js/romo/keypress.js
+++ b/assets/js/romo/keypress.js
@@ -1,0 +1,35 @@
+$.fn.romoKeypress = function() {
+  return $.map(this, function(element) {
+    return new RomoKeypress(element);
+  });
+}
+
+var RomoKeypress = function(element) {
+  this.elem = $(element);
+  this.defaultTriggerOn = 'keydown';
+
+  this.doInit();
+
+  this.triggerOn = this.elem.data('romo-keypress-on') || this.defaultTriggerOn;
+  this.elem.on(this.triggerOn, $.proxy(this.onTrigger, this));
+
+  this.elem.trigger('keypress:ready', [this]);
+}
+
+RomoKeypress.prototype.doInit = function() {
+  // override as needed
+}
+
+RomoKeypress.prototype.onTrigger = function(e) {
+  if (this.elem.hasClass('disabled') === false) {
+    this.doTrigger(e);
+  }
+}
+
+RomoKeypress.prototype.doTrigger = function(triggerEvent) {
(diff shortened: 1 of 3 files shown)


assets/js/romo/keypress.js
```diff
@@ -0,0 +1,35 @@
...
+  return $.map(this, function(element) {
+    return new RomoKeypress(element);
+  });
+}
+
+var RomoKeypress = function(element) {
+  this.elem = $(element);
+  this.defaultTriggerOn = 'keydown';
+
+  this.doInit();
```
jcredding, line 11: @kellyredding - I think you want to define an empty `doInit` function for this.

assets/js/romo/keypress.js
```diff
@@ -0,0 +1,35 @@
+$.fn.romoKeypress = function() {
+  return $.map(this, function(element) {
+    return new RomoKeypress(element);
+  });
+}
+
+var RomoKeypress = function(element) {
+  this.elem = $(element);
+  this.defaultTriggerOn = 'keydown';
```
jcredding, line 9: @kellyredding - We discussed maybe making this `keypress`. I can't remember all the details for each of the events but it might make more sense since the component is called keypress as well.

jcredding:
@kellyredding - A couple comments, I'm good with the event name, the html attribute names and the overall structure. Address the comments and I'm good with this :boom:

assets/js/romo/keypress.js
kellyredding, line 9: @jcredding I'm going to keep this as 'keydown' for now.  'keypress' is only triggered on character inputs (but not for modifiers).  I think it is better to detect all keys by default.

assets/js/romo/keypress.js
kellyredding, line 11: @jcredding thanks - done.

kellyredding:
@jcredding I also added this to the default dassets combination - I forgot that when I originally submitted this PR.

kellyredding:
@jcredding I've tested this in one of our apps and it works as intended.  I'm going to go ahead and merge, FYI.

PR #10, opened 2015-02-09
form js: update to queue submits to avoid two submits running at once

kellyredding:
This adds in submit queueing logic to the forms.  This is designed to
prevent a form from being submitted if it is currently running a
previous submission.

This is prep for having keypress inputs submit the form.  These
inputs will trigger submissions as the user types.  We don't want
to send a submission for every key entered.  It is better to queue
them up and send one once the prev submit is done.

@jcredding ready for review.  Again, check my design and handling on this.


jcredding:
@kellyredding - I'm good with the recursive design here and the changes make sense to me :boom:

PR #11, opened 2015-02-09
form js: detect keypress submit elems and submit on keypress trigger

kellyredding:
This updates forms to detect any elements that have been marked as
keypress submit elements.  It binds the keypress trigger event to
submitting the form.

Note: this adds a minimal delay timeout on the event binding.  This
avoids submitting the form on _every_ keypress.  Instead this behaves
more like "submit the form when there is any pause in typing".

@jcredding ready for review.  Feedback on the timeout value.  I just tried to choose the minimal value that didn't spam-submit as you typed.


jcredding:
@kellyredding - I like how relatively simple this was to add in and how it fits in. I don't have much feedback on the timeout value, I think I remember seeing values like 300ms and 500ms when we've implemented this kind of behavior in the past. All that to say I'm good with this :boom:

PR #12, opened 2015-02-10
onkey js: rename old "keypress" component to "onkey"

kellyredding:
This is just to avoid confusion with thinking Onkey works with only
the "keypress" event.  Onkey can be used with any "key" event and I
wanted to name this more generically.

There are no behavior changes here - just renames.

@jcredding ready for review.


jcredding:

PR #13, opened 2015-02-10
form js: configure onkey submit delay

kellyredding:
This updates the form js to first look for a submit delay data option
on the onkey submit elem, then fall back to the default sumbit delay.
This allows different onkey inputs to configure how long they pause
before submitting the form.

@jcredding ready for review.


jcredding:
@kellyredding - Looks good :boom: I like the name for the key, goes with `romo-form-onkey-submit`.

PR #22, opened 2015-04-15
update select and datepicker js to properly bind hover behavior

kellyredding:
The 'hover' event name is deprecated in jQuery.  This switches to
the more formal (and more appropriate) `mouseenter` event name for
binding the behavior.

I discovered this when I upgraded jQuery on an app and suddenly
select value hovering broke.

@jcredding ready for review.


jcredding:

PR #23, opened 2015-04-15
update base init helpers to manually init every elem in the given collection

kellyredding:
When performing the init operation on multi-elem collections, only
one elem in the collection would actually get the auto-init behavior
triggered.  This updates to manually do both the operation and the
auto-init trigger on all given elements manually to ensure it happens
consistently to each.

As a side-effect, these methods no longer return the elements they
operate on.  They now return nothing.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -29,46 +29,52 @@
     return $(elems);
   }
 
-  Romo.prototype.initHtml = function(elem, data) {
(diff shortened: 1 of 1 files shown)


jcredding:
@kellyredding - I'm good with this :boom: Since this no longer returns the elements I think we will have to update a few places but I think we can just save off what we are passing to the init methods which is the same as what it previously returned.

PR #24, opened 2015-04-15
add style classes to set to full width/height

### LINKED ISSUE #15: add generic class for width/height 100%
would be nice to use declaratively.  generic tables, for instance, don't get width 100%.  romo-table adds it but if you don't want to use romo-table, you are stuck.  this would allow generic tables with width 100%.

kellyredding:
This is a helper for setting width/height to 100%.  Handy for elems
that aren't 100% by default but need to be in special cases.

Closes #15.

@jcredding ready for review.

### DIFF
--- a/assets/css/romo/base.scss
+++ b/assets/css/romo/base.scss
@@ -427,6 +427,8 @@ h3 { @include text1; }
 
 /* other helpers */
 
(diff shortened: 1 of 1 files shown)


jcredding:

PR #27, opened 2015-04-15
add style class for centering using margin

### LINKED ISSUE #21: add style class for center align by margin
what to call it??  would set `margin: 0 auto`

kellyredding:
This only works with fixed width elems, but can be used as an
alternate for `romo-align-center` that won't affect text alignment.

Closes #21.

@jcredding ready for review.


jcredding:

PR #29, opened 2015-04-15
fixed table layout style class

### LINKED ISSUE #17: add style class for table layout fixed
maybe also a macro class for table layouts that does fixed plus width 100%??

kellyredding:
This adds a style class to force table layout fixed.  This is used
when you don't want the size of the table cols to adjust to their
content size.

Closes #17.

tables romo 2015-04-15 15-55-06

@jcredding ready for review.


jcredding:

PR #32, opened 2015-05-20
css: add style classes for vertically aligning based on text size

kellyredding:
These are similar to the `{button|label}-inline*` style classes, just
for text.  These are needed when you want to vertically align pieces
of text that aren't siblings in a parent element.

@jcredding ready for review.


jcredding:

PR #35, opened 2015-05-26
css: add style classes for grab-related cursors

kellyredding:
This is to compliment the existin pointer cursor style class.  This
is handy when doing drag related behavior and want to give cursor
feedback.

@jcredding ready for review.


jcredding:

PR #36, opened 2015-06-10
Add browser submit to forms

jcredding:
This updates forms to allow submitting using the browser instead
of Romos custom logic. This will call the `submit` method on the
form element and trigger standard form behavior. This is an option
that allows mixing standard form behavior (`target` attribute,
non-ajax request) with other Romo form features (submit on change,
loading indicator, form event).

Forms will now look for a browser submit option and, if set to
true, `submit` will be called on the form. Since this is submitting
the form, many of the events that are triggered will not be. The
form cannot tell if the submit was successful, it failed or when
it even completes. Instead of these events, the form will trigger
a browser submit event after it calls `submit`. This can be used
to bind behavior "after submit".

Finally, this proxies the new browser submit event through the
modal and dropdown form components.


kellyredding:
@jcredding this great - love how it turned out.  Also the "browser" naming scheme is holding up the more I look at it. :-1:

PR #38, opened 2015-06-15
select js: add set value method and trigger event

kellyredding:
This provides an API for programmatically setting select values.
This should be thought of as the equivalent of `.value =` on a
typical select elem.  Just like `.value =`, this _does not_ trigger
any "change" events.

@jcredding ready for review.


jcredding:

PR #39, opened 2015-07-02
form js: add an elem data option to disable keypress behavior

kellyredding:
The keypress behavior may not always be desired or it may be handled
by another set of logic.  This allows you to disable it on a per-form
basis.

Specifically this came up when integrating a romo form with a wysiwyg
and hitting 'Enter' in the wysiwyg caused the form to submit instead
of add a newline (oops).

@jcredding ready for review.


jcredding:
@kellyredding - Cool :boom:

PR #41, opened 2015-09-18
css: add 'base' and 'alt' text/bg colors

kellyredding:
This is in addition to the 'muted', 'warning', etc colors.  This
allows you to get to back to the base/alt colors when some parent
elem has some custom color set.  This is also useful to get back
to the base/alt on hover when non-hover has a custom color.

Check this for an example on tables:
image-hae0l

@jcredding ready for review.

### DIFF
--- a/assets/css/romo/_mixins.scss
+++ b/assets/css/romo/_mixins.scss
@@ -139,15 +139,23 @@
 
 /* color emphasis */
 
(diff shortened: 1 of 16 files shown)


jcredding:
@kellyredding - Looks good :boom: I like making the base and alt consistent with the other colors.

PR #42, opened 2015-09-18
css: border-radius-circle

kellyredding:
This takes the border radius logic and provides it as a general
border radius class.  Use this when you want to make any elem
look like a circle.

@jcredding ready for review.


jcredding:

PR #43, opened 2015-09-30
modal/dropdown: add API to bind/unbind body event handling

kellyredding:
Both modals and dropdowns add event binding on the window body
to close their popups when the body is clicked or the esc key
is hit.

This reworks the binding/unbinding into a formal API so that it
can be turned on/off.  This is specifically needed for the new
romo-av modal/dropdown video components.  Because of browser quirks
when videos are in fullscreen, modal/dropdown videos need to disable
the body events while in fullscreen.  The net effect is you can
hit esc to exit fullscreen and not dismiss the video modal popup.

In addition, this bind element key up handling on the modal.  This
was added out of necessity for the dropdown component which was
implemented after the modal was.  The modals body key up handling
covered the modal but I chose to add the elem key up handling to
keep it consistent with the dropdown component.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/dropdown.js
+++ b/assets/js/romo/dropdown.js
@@ -45,8 +45,8 @@ var RomoDropdown = function(element) {
   this.elem.on('invoke:loadError', $.proxy(function(e, xhr, invoke) {
     this.doLoadBodyError(xhr);
   }, this));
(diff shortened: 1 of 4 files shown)


jcredding:
@kellyredding - Looks good :boom: I like keeping the modal and dropdown consistent.

PR #44, opened 2015-10-12
indicator js: capture css/html contents on start

kellyredding:
This as opposed to on init.  Sometimes the full css width/height
etc aren't set at init-time (due to markup being lazy loaded, etc).
This allows the dimensions and markup to change and the captured
css/html to handle this appropriately.

This switches to conditionally setting the html on stop.  This case
has to be handled now that the html isn't being captured on init.

Finally, this switches to capturing the style attr it overrides
and setting it on stop.  This ensures it safely applies the css
necessary while the indicator is running but resets the state and
cleans things up appropriately.

@jcredding ready for review.  I think you and I have both seen issues with this when using font awesome icons in buttons that are indicators, for example.

### DIFF
--- a/assets/js/romo/indicator.js
+++ b/assets/js/romo/indicator.js
@@ -31,12 +31,6 @@ var RomoIndicator = function(element) {
   this.doInit();
   this.spinner = new Spinner(this.spinnerOpts);
 
-  this.elemHtml = this.elem.html();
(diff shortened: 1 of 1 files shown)


jcredding:
@kellyredding - Yeah, I've run into this one before. Looks good :boom:

PR #45, opened 2015-10-12
css: add style class for color text the disabled color

kellyredding:
This disabled color styling is already used in the disabled buttons,
disabled dates in the datepicker and disabled options in the selects.
This adds a style class so users can manually change to this color
in ad-hoc markup.

@jcredding ready for review.


jcredding:

PR #46, opened 2015-10-12
tooltip js: allow changing content data; set content API

kellyredding:
The prev implementation saved off the content data attr at init-time.
This meant that if the attr changed, it wouldn't be honored when
showing the tooltip.

This switches to lazily grabbing the data attr on popup open.  This
allows the attr to change and have the current attr always be displayed
in the tooltip.

This also adds a methods and a trigger event for setting the tooltip
content.  This will allows the tooltip content to be updated while
the tooltip popup is open.

@jcredding ready for review.


jcredding:

PR #47, opened 2015-10-12
modal/dropdown/tooltip js: api to specify custom append to elems

kellyredding:
By default, the popups for these components are appended to the
body.  This is a safe default as it will minimize the number of
parent styles that get trickled down to your popups.  However if,
for example, you want to full screen some markup that uses these
components you won't be able to see their popups by default (b/c
the body isn't being full screened).

This adds a data attribute for specifying which elem to append the
popup to.  It will still default to the body, but can be passes a
selector.  The selector will be used to find the closest matching
elem and append the popup to that elem.

@jcredding ready for review.


jcredding:

PR #48, opened 2015-10-14
dropdown: turn off body event handling if modal is opened

kellyredding:
This updates the dropdown component to turn off its body event
handling while a modal (that was opened from the dropdown) is open.
The effect here is that any body event handling will be limited to
the modal while it is open and the dropdown will ignore body event
handling until the modal is closed.

This only applies to modals opened from dropdowns.  Modals opened
outside of the dropdown behave as before.

Note: this also switches dropdowns/tooltips to close on modal _moves_.
The old logic closed all popups as soon as you mouse down.  This is 
technically not necessary as problems don't come up until you
start moving the modal (dragging it).  The changes to wait as long
as possible before closing.  This was noticed while messing around
with the above changes.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/dropdown.js
+++ b/assets/js/romo/dropdown.js
@@ -61,6 +61,16 @@ RomoDropdown.prototype.doInit = function() {
 RomoDropdown.prototype.doInitBody = function() {
   this.doResetBody();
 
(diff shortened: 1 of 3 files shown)


jcredding:
@kellyredding - I'm cool with this :boom:

PR #50, opened 2015-10-28
invoke js: add options method data option

kellyredding:
This allows you to use a different ajax method for loading data
but still default to 'GET' if no custom method is supplied.  This
allows you to use invokes in more applications than just simply
fetching data with a 'GET'.

@jcredding ready for review.


jcredding:

PR #51, opened 2015-11-03
js: track parent-child elem relationships; use to remove appropriately

kellyredding:
The problem we are tying to solve here is that some js components
that bind to DOM elems need to create additional DOM elems that
are not hierarchial children but are in their nature/design.  These
generated DOM elems are useless without the source elem.  However,
if the source elem was removed, these DOM elems would linger b/c
they existed in a separate elem tree.

This adds a parent/child elem tracking class that monitors DOM
removals at the body level.  If it detects a node removal that
contains parent elems, it goes and looks up the child elems for
each parent and removes them.  This keeps the DOM clean and avoids
weird edge-case behavior of orphaned DOM elements lingering when
they shouldn't.

This has been added to the tooltip, modal, dropdown, select and
datepicker components.  The tooltips, modals and dropdowns all
build popup elems that are appended to the body.  The selects appends
a wrapper after itself while the datepicker appends a wrapper and
then nests itself in said wrapper.  These wrappers/popups should
be removed if/when their component elems are removed.

The tracking class is a collection with `add` and `remove` methods.
Components use the `add` method to associate child elems with parent
elems.  The `remove` method is used by the DOM observer to remove
all child elems for parent elems in a node's tree.  Internally, the
tracking class generates elem ids and sets them for usage as a data
attr on the parent elems.

This gif shows two scenarios: 1) me replacing a set of markup that has tooltips and a dropdown to show that elems are no longer just accumulating and 2) a case where I use javascript to remove all parent elems and you see that the generated markup is also removed (ie all the popups).

gif

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -4,6 +4,8 @@
   }
 
   Romo.prototype.doInit = function() {
+    this.parentChildElems = new RomoParentChildElems();
+
     $.each(this._eventCallbacks, function(idx, eventCallback) {
       $('body').on(eventCallback.eventName, eventCallback.callback);
     });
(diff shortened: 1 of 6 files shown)


jcredding:
@kellyredding - Looks good :boom: I like moving the popup stuff into `doInitPopup` methods. 

The parent child elems is cool and I think is well designed for working in the constraints of browser environment. I can't think of a better way to fix the problem and I think all the negatives of this implementation are handled as best as possible (trying to exit out of the observer as quickly as possible, removing items from the hash with the pop, wrapping all the logic behind a single class and simple api).

PR #52, opened 2016-01-10
Make `Dassets.configure!` safe to call multiple times

jcredding:
This makes the `Dassets.configure!` method safe to call multiple
times by setting a flag to true when its called. It checks the flag
when its called and if its true it returns doing nothing. This is
so multiple gems can use romo and configure it with dassets without
being aware of one another or worrying about call order.


kellyredding:
@jcredding word man - :-1:

PR #53, opened 2016-01-21
form js: add ability to confirm form submissions

kellyredding:
To enable this, configure your submit elem `data-romo-form-submit`
attr with "confirm" instead of the standard "true" value.  This
will trigger the `form:confirmSubmit` event instead of submitting
the form.  Use this event to do whatever you need to confirm the
form submission.  Once, confirmed, manually call `doSubmit` on the
form object or alternatively trigger the `form:triggerSubmit` event
on the form elem and the form will submit as usual.

This event has also been setup to proxy from dropdown, modal and
inline form components.

@jcredding ready for review.


jcredding:

PR #54, opened 2016-01-21
invoke js: queue invoke requests like forms queue submissions

kellyredding:
I noticed that forms queue their ajax submission requests however
invokes don't queue their ajax requests.  There really isn't a
good reason for this.  So, this adds request queueing to invokes
to ensure that only 1 invoke is ever run at one time.  Any attempts
to invoke while a current invoke is in progress will be run once
the current invoke is finished.

@jcredding ready for review.


jcredding:

PR #57, opened 2016-02-09
Update inline form to propagate the inline show event

jcredding:
This updates the inline form to propagate the inline show event.
This is the pattern we use for all the other inline events. When
the event was added we forgot to update the inline form. This
allows javascript logic to only bind to inline form events and
not have to know anything about the inline it composes.

This also removes propagating the inline invoke event. The event
was removed and is no longer published by inline components.

Related to #55 (662d7b1)


kellyredding:
@jcredding looks good to me man :-1:

PR #58, opened 2016-02-16
datepicker: add trigger events to set format/date

kellyredding:
This allows you to access these functions via events (for cases where
you don't have access to the auto-initialized component instance).
This follows the patter of trigger events on this and other components.

@jcredding ready for review.


kellyredding:
@jcredding going to go ahead and merge this since you are out right now.

PR #59, opened 2016-02-16
always trigger datepicker change events on elem change

kellyredding:
This adds some consistency to the events that are published on
datepickers.  Before, a 'datepicker:change' event would only fire
when a user selected an date from the popup.  If they manually typed
a new date in, the event wouldn't fire.  This reason is b/c we had
no way to provide the previous value needed for the event.

This switches to manually tracking a previous value and to always
publish a datepicker change event on every underlying elem change
event.  Prefer binding to 'datepicker:change' over 'change' and
get access to the new/prev values plus the datepicker component.
This datepicker change event will now fire in the following  date
change scenarios: date selected in popup, set date event trigger,
and manual user edit.

The only date change scenario that isn't covered is when the `doSetDate`
method is manually called.  The reason is that this is a low-level API
that is used to compose other behavior.  If you are changing the date
via this method, you are responsible for triggering a 'change' event
if needed (which will in turn trigger the datepicker change event.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/datepicker.js
+++ b/assets/js/romo/datepicker.js
@@ -11,13 +11,14 @@ var RomoDatepicker = function(element) {
     "January", "February", "March", "April", "May", "June",
     "July", "August", "September", "October", "November", "December"
   ]
-  this.defaultPrevClass = undefined;
-  this.defaultNextClass = undefined;
-  this.defaultIndicatorClass  = undefined;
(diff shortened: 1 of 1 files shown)


kellyredding:
@jcredding I've tested all this out in our big app and it is needed for a new feature.  I'm going to go ahead and roll this since you are out today.

PR #60, opened 2016-02-19
datepicker: add API to show/hide datepickers

kellyredding:
Special handling is needed due to the elem wrapper and the indicator
elem.  This follows the pattern of the enable/disable API.

Note: I had to use `css` instead of `show`/`hide`.  Show sets
`display:block` which breaks the indicator element UI so I want to
be safe and just remove the explicit display setting.  I also added
display inline block to the indicator's formal styles since it is
needed.

@jcredding ready for review.


jcredding:

PR #61, opened 2016-03-18
css: make color CSS most important for table components

kellyredding:
This applies to both regular and grid tables.  The idea is that
a row color class should take precedent over any striping classes
that might be in play.  This came up when using a color class on
a hovered striped grid table.  It would apply but would be overridden
when hovering b/c the striped hover css took precedent.

This adds `!important` to ensure it takes precedent.

@jcredding ready for review.


jcredding:

PR #62, opened 2016-03-24
selects/datepickers: better handle dynamic carets/indicators

kellyredding:
We allow you to set custom caret/indicator style classes (which in
turn have custom widths) but we were hard-coding `22px` of padding
to make room for the caret/indicator UI.  This is incosistent and
brittle as it is too easy to "cut off" the UI b/c the custom class
specified UI that is wider than 22px.

This adds additional API to both components that allows specifying
a width in pixels in additon to the style class.  Romo now calculates
the necessary padding to show the UI based on this custom value
(accounting for padding on the left/right sides).

Note: the "1px optical illusion I reference in the select code is
due to how Romo renders selects.  The select UI has a border that
is not at the true right edge of the element - there is an additional
1px "shadow".  The shadow is the actual edge of the element.  However,
our eyes see the border as the edge of the element so we need to
account for this in the padding/positioning calculations.

Note: this also moves defaulting the select caret values to the 
select js file.  It was previously in the select dropdown file by
mistake.

@jcredding ready for review

@Criptak FYI

### DIFF
--- a/assets/js/romo/datepicker.js
+++ b/assets/js/romo/datepicker.js
@@ -11,14 +11,15 @@ var RomoDatepicker = function(element) {
     "January", "February", "March", "April", "May", "June",
     "July", "August", "September", "October", "November", "December"
   ]
-  this.defaultPrevClass      = undefined;
-  this.defaultNextClass      = undefined;
-  this.defaultIndicatorClass = undefined;
-  this.itemSelector          = 'TD.romo-datepicker-day:not(.disabled)';
-  this.calTable              = $();
-  this.date                  = undefined;
-  this.today                 = new Date;
-  this.prevValue             = undefined;
+  this.defaultPrevClass        = undefined;
+  this.defaultNextClass        = undefined;
+  this.defaultIndicatorClass   = undefined;
+  this.defaultIndicatorWidthPx = 0;
+  this.itemSelector            = 'TD.romo-datepicker-day:not(.disabled)';
+  this.calTable                = $();
+  this.date                    = undefined;
+  this.today                   = new Date;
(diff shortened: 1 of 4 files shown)


assets/js/romo/datepicker.js
```diff
@@ -67,7 +68,13 @@ RomoDatepicker.prototype.doBindElem = function() {
       this._hide(this.indicatorElem);
     }
     this.indicatorElem.on('click', $.proxy(this.onIndicatorClick, this));
-    this.elem.css({'padding-right': '22px'});
+
+    var indicatorWidthPx   = this.elem.data('romo-datepicker-indicator-width-px') || this.defaultIndicatorWidthPx;
+    // left-side spacing
+    // + caret width
```
jcredding: @kellyredding - minor, indicator instead of caret to line up with the variable names

assets/js/romo/select_dropdown.js
```diff
@@ -6,7 +6,8 @@ $.fn.romoSelectDropdown = function(optionElemsParent) {
 
 var RomoSelectDropdown = function(element, optionElemsParent) {
   this.elem = $(element);
-  this.defaultCaretClass = '';
+  this.defaultCaretClass   = undefined;
+  this.defaultCaretWidthPx = 0;
```
jcredding: @kellyredding - Shouldn't these be on the select component and not the select dropdown? In the select js diff, we use `this.defaultCaretClass` and `this.defaultCaretWidthPx` and `this` seems to reference an instance of a select component not a select dropdown component.

jcredding:

PR #63, opened 2016-06-10
get working in modern ruby versions

kellyredding:
This updates the gem dependencies to bring in the latest versions
(which also now work in modern ruby versions) and updates the test
suite to get up to convention for working in modern ruby versions.
This includes backfilling Array#sample so it works in 1.8.7.

Note: I also added a test support factory and switched to "vanity
versioning" in the Gemfile.  Both of these changes just get the
code up to our latest conventions.

@jcredding ready for review.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,12 +5,12 @@ gemspec
 gem 'pry', "~> 0.9.0"
 
 gem 'rack'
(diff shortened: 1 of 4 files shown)


jcredding:

PR #65, opened 2016-10-25
datepicker: honor timezone when determining today's date

kellyredding:
This updates how today's date is calculated to properly honor
time zones.  Datepicker uses UTC dates for comparisons.  Before
we were assuming `new Date()` was a UTC date for today when in
fact it was a time zoned date.  This uses the timezoned date info
to build a UTC date for the date as perceived in the current time
zone.  Then this new UTC date is used for the "today" comparisons.

romo-datepicker-today-fix

@jcredding ready for review.  I noticed this while using the datepicker late at night.  Noticed the screenshot with the simulated late time but the correct "today" date.


jcredding:
@kellyredding - Looks good and nice catch :boom:

PR #67, opened 2016-10-25
tooltip: have tooltips intelligently choose top vs bottom pos

kellyredding:
This is similar behavior to the dropdowns.  If a tooltip is set
to position top or bottom, it will first check to make sure there
is enough room to position itself that way.  If not, it will
switch to the opposite pos as it is assumed there will be enough
room that way.

The only wrinkle that is different from dropdowns is the tooltip
arrow UI.  I had to unbind the CSS for positioning the arrow from
the element configured position setting.  It now needs to be based
on a specific arrow position setting we set on the popup once we
determine the live position of the popup.

This only applies to top/bottom positioning.  We chose not to
implement auto left/right positioning as it comes up so less
frequently and can be designed around in almost all cases.  We
may choose to add it later as specific needs arise.  This all is
true for dropdowns.

@jcredding ready for review.

### DIFF
--- a/assets/css/romo/tooltip.scss
+++ b/assets/css/romo/tooltip.scss
@@ -30,28 +30,28 @@
   border-style: solid;
 }
 
-.romo-tooltip-popup[data-romo-tooltip-position="top"] .romo-tooltip-arrow {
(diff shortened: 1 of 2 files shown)


jcredding:

PR #68, opened 2016-10-26
use `getComputedStyle` to lookup/parse elem z-index values

kellyredding:
This is used to set appropriate cumulative z-index values for
modals/dropdowns/etc.  Chrome's latest update broke our old way
of getting an elem's z-index.  This switches to the more robust
and now fully supported `getComputedStyle` which works in all
browsers.

Note: I had to switch to ignoring `0` index values as this is the
default value of `getComputedStyle`.  Since we add zero on in the
end, the default effect is the same.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -149,24 +149,32 @@
   // style handling
 
   Romo.prototype.parseZIndex = function(elem) {
-    // for the case where the browser doesn't suck and can read inherited z-index
-    var val = parseInt(elem.css('z-index'));
-    if (!isNaN(val)) {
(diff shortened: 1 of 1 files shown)


kellyredding:
@jcredding FYI, see this for reference:
- http://stackoverflow.com/questions/1118198/how-can-you-figure-out-the-highest-z-index-in-your-document
- http://erik.eae.net/archives/2007/07/27/18.54.15/ (the IE in-compatibility is out of date)
- https://msdn.microsoft.com/en-us/library/ff975168(v=vs.85).aspx

jcredding:

PR #69, opened 2016-10-27
add an IndicatorTextInput js component and style classes

kellyredding:
This was extracted from the DatePicker component and commonizes
creating text inputs with some sort of "purpose indicator" icon.
This handles the positioning of the icon relative to the text
input and triggers an event when the indicator is clicked.

This will be used in some future efforts:
- possibly rework date picker to compose this
- add search inputs to select dropdowns
- a new "picker" component that is similar to selects but options
  aren't pre-loaded

@jcredding ready for review.


assets/js/romo/indicator_text_input.js
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  this.elem.before(elemWrapper);
+  elemWrapper.append(this.elem);
+
+  // the elem wrapper should be treated like a child elem.  add it to Romo's
+  // parent-child elems so it will be removed when the elem (input) is removed.
+  // delay adding it b/c the `append` statement above is not a "move", it is
+  // a "remove" and "add" so if added immediately the "remove" part will
+  // incorrectly remove the wrapper.  Any value will do - I chose 100 arbitrarily.
```
jcredding: @kellyredding - I don't think you are using 100 below

jcredding (approved):

PR #70, opened 2016-10-28
add a `toArray` helper global function

kellyredding:
This commonizes a couple places where we've manually converted a
jQuery-style element collection to a vanilla js array of nodes.
I chose to implement this function in pure js instead of relying
on the jQuery-specific `toArray` method on collections.  I am
trying to move to using vanilla js whenever possible as we are
targeting modern browsers and they are getting pretty good these
days.

This will also be needed in a coming effort to implement a method
to do word-boundary filtering on collections of elements.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -156,9 +156,7 @@
     }
 
     // for the case where z-index is inherited from a parent elem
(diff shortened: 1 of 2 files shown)


jcredding (approved):

PR #71, opened 2016-10-29
add `getComputedStyle` global helper function

kellyredding:
This is just a more succinct way of calling the `window` function
with a friendly api and less boilerplate.  Now that I am trying
to use vanilla js in Romo, I need to calculate styles with this
method and having this helper makes it easier to do so.

@jcredding ready for review.


jcredding:

PR #72, opened 2016-10-31
add `RomoWordBoundaryFilter` class

kellyredding:
This class filters the given `setElems` by a given `filterString`.
The filter groups the set elems into two groups: ones that match
the filter string vs ones that do not match filter string.

The filter matches based on matching "word boundaries", meaning
each word in the filter string needs to match the beginning of a
word in the elem's text content the words must match in the same
order as the filter string.  The word boundaries are anything
that matches `/[\s-_]+/` for now.  This means hyphenated words
and underscored words are treated like space-separated words
which provides more flexibility to the filter without
compromising its "word boundary" nature.

This will be used in a future effort to provide selects with an
input to filter the select options as you type.

@jcredding ready for review.


jcredding:

PR #73, opened 2016-10-31
add `data-romo-form-disable-enter-submit` option to forms/inputs

kellyredding:
This allows users to turn off the default "hit enter to submit the
form" behavior both form-wide and on a per-input basis.  This
came up in an app we have where there is a form with an input but
the input is used to just filter the checkboxes shown and should
not be used to submit the form.

@jcredding ready for review.


jcredding:

PR #74, opened 2016-11-07
place indicator text input indicator on demand

kellyredding:
This adds a method and event trigger to re-place the indicator
for indicator text inputs on demand.  This allows you to re-run
the logic that calculates where to place the indicator UI as
needed (typically do to resizing the input after the logic
initially runs.

Specifically, this is needed for a future effort where we add
filter indicator text inputs to select dropdowns.

@jcredding ready for review.


jcredding:

PR #75, opened 2016-11-07
refine dropdown placement logic

kellyredding:
This switches to rounding the top/left pixel values to the nearest
100th place value.  This prevents "jitter" as you call to place
the popup as repeated calls with overly precise values will not
necessarily calculate the same overly precise value on subsequent
calcs.  This occured when I had the new select option filters call
to place the dropdown popup each time the options were filtered.
In some cases the popup would "jitter" and slightly update its
position as I was pressing keys to filter the results.  Using
less precise values sacrifices nothing on position accuracy while
making the placement values deterministic.

This also updates the placement calcs to account for the popup
and content elems not being the same height.  This is also prep
for adding in option filter UI to the select dropdowns.  This UI
will be adding to the popup but not in the content area.  The
content gets max-height added so that overflow content can be
scrolled.  This filter UI should never be overflowed and should
always be visible.  This updates the calcs to account for any
elem height differences and builds that into the overflow height
handling logic.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/dropdown.js
+++ b/assets/js/romo/dropdown.js
@@ -333,7 +333,6 @@ RomoDropdown.prototype.onResizeWindow = function(e) {
 RomoDropdown.prototype.doPlacePopupElem = function() {
   if (this.elem.parents('.romo-modal-popup').size() !== 0) {
     this.popupElem.css({'position': 'fixed'});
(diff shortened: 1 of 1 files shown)


jcredding:

PR #76, opened 2016-11-07
refine select next/prev option logic

kellyredding:
This is prep for adding a filter for select options.  When testing
the filter in a grouped select, if you filtered such that some of
the groups were empty, it broke cycling to the prev or next opt.

This reworks the logic to both be more simple and easy to read,
but also work with empty nested option lists.  This new
implementation includes adding `selectPrev` and `selectNext`
Romo utility helper methods.  These behave like `$().prev` and
`$().next` but take a selector and only consider elems that
match the selector.  These are used to cycle to the next/prev
opts in the current list and if there is no opt in the current
list, then cycle to the prev/next list.

Note, this also includes a few minor handling tweaks that are
somewhat related:

* this disables the `tab` key while the popup is open.  this
  matches how native selects seem to work so I wanted to emulate
  it in Romo's select
* opens the popup on the 'Up' key as well as the 'Down' key.
  again, this is how the native selects seem to work so I want
  Romo's to emulate this.

Without the fix:
select-next-prev-fix

And here with the fix:
select-next-prev-with-fix

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -182,8 +182,34 @@
   // elem handling
 
   Romo.prototype.toArray = function(elems) {
-    // converts a collection of elements (`$()`) to an array of nodes
+    // converts a collection of elements `$()` to an array of nodes
     return $.map(elems, function(node){ return node; })
+  }
+
+  Romo.prototype.selectNext = function(elem, selector) {
+    // like `$().next()`, but takes a selector; returns first next elem that
+    // matches the given selector or an empty collection if non matches
+    var el = elem.next();
+    while(el.length) {
+      if (selector === undefined || el.is(selector)) {
+        return el;
+      }
+      el = el.next();
+    }
+    return el;
+  }
+
+  Romo.prototype.selectPrev = function(elem, selector) {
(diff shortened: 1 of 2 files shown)


assets/js/romo/base.js
```diff
@@ -182,10 +182,36 @@
...
+      }
+      el = el.next();
+    }
+    return el;
+  }
+
+  Romo.prototype.selectPrev = function(elem, selector) {
+    // like `$().prev()`, but takes a selector; returns first prev elem that
+    // matches the given selector or an empty collection if non matches
+    var el = elem.prev();
```
jcredding, line 205: @kellyredding - I think jquery's `prev` and `next` take a selector, but if you are wanting to get away from depending on that logic I'm cool with having this.

jcredding (approved):
@kellyredding - I like the simplifications, looks good :boom:

assets/js/romo/base.js
kellyredding: @jcredding dang, I misread their docs and thought they didn't take selectors.  Oh well, the eventual goal is to not require jQuery so I'll leave this as is for now.

PR #77, opened 2016-11-07
make select options filterable

kellyredding:
This adds a RomoIndicatorTextInput for filtering select options
in the select dropdown.  This allows you to type on selects when
focused and get presented with a filtered set of options matching
what you typed.  You can then cycle the filtered set and select
one as you would on the full set.

This is intended to be an enhancement on standard select handling
when you type in them.  By default, selects allow you to type
on them and select the first elem that matches what you type.
There are a few disadvantages though:

* you don't see what you've typed
* you have to type in the option value exactly
* it only highlights the first option matching what you typed -
  you don't have any idea of how many other things match as well
* to reset you have to wait an unknown amount of time and know to
  start over fresh from the beginning

The improves on that default behavior in a few ways.  First, you
have a separate text input embedded in the select's dropdown so
you see what you've typed and can intuitively change that or
clear it on demand.  Second, it filters the options removing any
that don't match what you've typed.  This allows you to see all
matching options and cycle over them like you would the full set.
Third, the filter uses the RomoWordBoundaryFilter to filter the
options.  This allows more friendly filtering as you don't have
to type the full option names and instead just quickly type the
beginnings of option value words.

One big concern here is how to handle the filter input especially
for selects in forms.  This adds extra css and js logic to make
the select still *seem* focused while the filter input is focused
and css to emulate the text input styles for the option filter.
The option filter also has no `name` attr so it won't be submitted
in forms.  There is logic to properly re-focsus the select when the
dropdown is closed and/or an option is selected, including a new
dropdown event fired when the Esc key is pressed specifically
so we know to re-focuse the select.

This adds a utility function to Romo defining what keycodes are
not input text.  This is used to determine if key input on the
selects should trigger opening the dropdown and filling in the
filter or not.  We want to auto filter when typing on selects but
not for control chars like tab, etc.

The filter is added to the popup outside of the popup content.
This is so the filter is always shown even if the content
overflows the popup available height.

Finally there is a settings API, both for Selects and
SelectDropdowns that allows you to customize the filter
placeholder text, the filter indicator, and to not use a filter.

gif

gif1

@jcredding ready for review.

### DIFF
--- a/assets/css/romo/_vars.scss
+++ b/assets/css/romo/_vars.scss
@@ -27,6 +27,8 @@ $spacingSize0: $spacingSize1 * 0.35;
 $spacingSize2: $spacingSize1 * 2;
 
 $inputHeight: ($baseLineHeight / 2) * 3; /* 30px */
+$inputFilterHeight: $baseLineHeight;     /* 20px */
+
 $inputSize1: 220px;
 $inputSize0: $inputSize1 - 120px;
 $inputSize2: $inputSize1 + 120px;
--- a/assets/css/romo/forms.scss
+++ b/assets/css/romo/forms.scss
@@ -26,7 +26,8 @@ label { display: block; }
 
 .romo-form input,
 .romo-form select,
-.romo-form textarea {
+.romo-form textarea,
+input.romo-select-dropdown-option-filter {
   font-size: 100%;
   @include rm-push;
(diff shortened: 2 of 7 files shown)


jcredding (approved):
@kellyredding - Very nice :boom:

PR #78, opened 2016-11-07
have RomoDatepicker compose RomoIndicatorTextInput

kellyredding:
The RomoIndicatorTextInput was extracted from RomoDatepicker
originally and has become its own sub-component (used in the
RomoSelectDropdown filter UI for example).  This completes the
extraction from RomoDatepicker by having the date picker use the
indicator text input component. This also proxies the necessary
settings api and event api.

This also handles a few smaller tweaks:

* the indicator text input no longer turns off autocomplete. this
  should never have been done b/c it is too low level for this
  type of setting.  the select option filter and datepicker
  manually set this option b/c they need it but we don't want to
  force it on all indicator text inputs.
* the indicator text input triggers an event when the indicator
  is clicked.  the datepicker needs to bind behavior to this event.
* the select and select dropdown proxy filter indicator width px
  value.  this allows specifying how much width the indicator ui
  takes to prevent having text flow behind the indicator ui.

@jcredding ready for review.  I've tested out the new datepicker in our apps, both in standard forms and in our timesheet app that places it in a toolbar row.  Everything seems to work as before.

### DIFF
--- a/assets/js/romo/datepicker.js
+++ b/assets/js/romo/datepicker.js
@@ -11,15 +11,13 @@ var RomoDatepicker = function(element) {
     "January", "February", "March", "April", "May", "June",
     "July", "August", "September", "October", "November", "December"
   ]
-  this.defaultPrevClass        = undefined;
-  this.defaultNextClass        = undefined;
-  this.defaultIndicatorClass   = undefined;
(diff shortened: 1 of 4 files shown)


jcredding:

PR #79, opened 2016-11-08
tweak sortable to allow `onInit` overrides to add markup settings

kellyredding:
This reorganized the init logic to break out a "bind draggables"
method that is run post `onInit`.  This delays grabbing the
markup settings values until after any user-custom init logic so
that logic is able to set markup settings if needed.

Note: this also includes a really minor datepicker tweak to better
organize default settings.  I noticed this cleanup while doing
this effort.

@jcredding ready for review.


jcredding:
@kellyredding - Cool by me :boom:

PR #80, opened 2016-12-27
add raw color style classes based on Tachyons colors

kellyredding:
This adds style classes for coloring the text/bg/border (both
always or hover-only) for a given set of human-named colors.  The
color set is based on the color set from Tachyons CSS toolkit (see
http://tachyons.io/#colors for reference).

The goal here is to color UI in an explicit manner instead of
relying on the more implicit "emphasis" color classes we already
have available.  The implicit are still needed as the color used
for each "emphasis" can be customized.  This just adds the explicit
couterpart for coloring.  In a future effort, I'm going to change
the implicit "emphasis" colors to be based on the closest matching
explicit color variable.  This will provide some consistency to
the color set Romo offers.

@jcredding ready for review.

### DIFF
--- a/assets/css/romo/_mixins.scss
+++ b/assets/css/romo/_mixins.scss
@@ -147,53 +147,155 @@
 
 @mixin text-disabled($i:"") { color: $disabledColor #{$i}; }
 
-@mixin text-base($i:"")       { color: $baseText #{$i}; }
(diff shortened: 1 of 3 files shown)


jcredding (approved):
@kellyredding - I'm good with this as a starting place and I like having explicit colors to use when the emphasis colors don't make sense. Good stuff :boom:

PR #81, opened 2016-12-28
rework colors to prep for using in romo "emphasis" styles

kellyredding:
The "emphasis" styles are warning, error, info, success.  In a
future effort, I want to switch our emphasis styles to be based on
the new explicit set of colors.  This adds some colors and tweaks
some existing ones to better prep for that transistion and better
match the previous existing emphasis colors.  Some notes:

* I renamed the prev "gold" color to "orange and the prev "orange"
  color to "darkOrange".  This follows the pattern on other colors
  better.  Also, I darkened the gold color so that it more looks
  like "orange" now as well.
* I added "pastel" variants for red, blue, green, and yellow. These
  are needed for out emphasis table bg colors and compliment the
  palette I think.  I also removed "lightestBlue" as it is very
  close to "pastelBlue" and I think the name is stupid.
* I moved the "washed" colors to be jux'd below their corresponding
  "pastel" color.  I felt it made more sense to group them this
  way.

@jcredding ready for review.


jcredding (approved):
@kellyredding - I'm good with this :boom: I like getting rid of "lightest" cause there were already a lot of blues and having a "lightest", "washed" and "pastel" seems a bit much.

PR #82, opened 2016-12-28
rework "emphasis" text/bg/hover/borders/buttons/labels

kellyredding:
This reworks these elements to be based off of our new color vars.
In general, "warning" is now based off of `$orange`, "info" off of
`$blue`, "error" off of `$red`, and "success" off of `$green`.
This also formally defines a color for "inverse" (black).

The goal of all of this is to minimize color variation in Romo's
default styles.  To help with this, I've also updated other elems
(like tabs and tooltips) to base off of defined colors.  I also
reworked the button handling to not "mix" the bg/highlight colors
and just match the color it was based on exactly.  I chose to
darken the emphasis text colors slightly to help with readability.
Note: I chose to darken the warning button less on highlight than
the other buttons.  This is bc the 10% looked too dark for that
orange color.

I chose to remove the "primary" button/label.  I've felt like these
should be just "info".  Also, they were the only things that had
a "primary" so were kind of outliers.

New:
after

Old and Busted:
before

@jcredding ready for review.

### DIFF
--- a/assets/css/romo/_mixins.scss
+++ b/assets/css/romo/_mixins.scss
@@ -360,18 +360,14 @@
 /* Buttons */
 /* ------- */
 
-@mixin button-bg($start, $end, $text) {
+@mixin button-bg($bg, $hover, $text) {
+  background-color: $bg;
   color: $text;
-  border-color: $end $end darken($end, 10%);
+  border-color: $hover $hover darken($hover, 10%);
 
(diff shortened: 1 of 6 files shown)


jcredding (approved):
@kellyredding - I'm good with this :boom: Like the simplification of the `button-bg` mixin and removing the primary stuff. Changes make sense and screenshot looks great.

PR #83, opened 2016-12-28
cleanups to the color-related css, var names, etc

kellyredding:
This is some tedious cleanup in prep for adding color classes
for the buttons, labels, tables, and grid-tables.  I did a bunch
of things here:

* renamed some "Background" vars to "Bg".  I wanted shorter var
  names bc less typing and also I want to reformat some styles to
  have more on a single line and be vertically aligned.
* vertically align button/label class defs related to emphasis
  colors.  It is easier to mass edit when vertically aligned. This
  will help when I go add twenty-something color-related classes
  here.
* removed and `-error` style classes; renamed to `-danger`.  This
  goes for vars too.  I want to simplify the API so if you want
  "red" something, use "danger".  I preferred "danger" over "error"
  bc it is more friendly and doesn't imply you did something wrong
  like "error" does.
* I did a bunch of style cleanups and tried to align vertically
  where possible.  This helps with reading and mass editing.

@jcredding ready for review.  You cool with me removing "-error" style classes in favor of "-danger".  It will be a pain to rollout to our apps but I like the API simplification.

### DIFF
--- a/assets/css/romo/_mixins.scss
+++ b/assets/css/romo/_mixins.scss
@@ -147,91 +147,91 @@
 
 @mixin text-disabled($i:"") { color: $disabledColor #{$i}; }
 
-@mixin text-base($i:"")       {            color: $baseText #{$i}; }
-@mixin text-base-hover($i:"") {            color: $baseTextHover #{$i}; }
-@mixin bg-base($i:"")         { background-color: $baseBackground #{$i}; }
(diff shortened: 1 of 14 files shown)


jcredding (approved):
@kellyredding - Looks good :boom: I like the consistency of just having danger everywhere.

PR #84, opened 2016-12-29
add missing text hover vars/mixins/classes; cleanup colors defs

kellyredding:
This is more prep work for adding our new colors to our CSS
components: buttons, labels, tables, and grid-tables.

This first reorganizes some things: the input, tab, and tooltip
color vars are now jux'd next to the base/alt/etc "basics" colors
in the vars file.  This not only keeps them organized with the
large number of color vars we've added, but also places them
below the explicit color vars so they can reference them.

This switches to using `$blue` for both the "input highligh" color
(ie for selects, datepickers, etc) so we limit how many "blue"
variants are in use.  This also switches the input colors to use
basic values, again, to limit the color variants in use.

When doing the docs for the colors, I noticed the pastel colors
and washed colors were very similar in appearance.  I chose to
keep the more "vibrant" of the two (the pastel version in all but
red) and call that "pastel" and then delete the "washed" colors.
I didn't see any value or keeping when they were so similar to
the pastel variants but more faint.  I preferred the "pastel" name
so I chose to keep that name over "washed".  I also noticed the
light/pastel yellows were *very* similar.  I chose to remove the
light yellow bc I wanted to keep the pastel yellow to match the
other pastel colors.  I darkened the pastel yellow slightly to
make up a bit for losing light yellow.  I also looked nice against
the other pastel colors.  I removed the light pink color b/c it
was so similar to both the pink and pastel red.

Finally, this adds explicit color text/bg hover vars/mixins. These
will be needed, not only to be consistent with emphasis color
handling, but also b/c the tables will reference them in a future
effort.  I updated the text/bg style classes to use these new
vars.  I also organized all color style classes, both emphasis and
explicit, into the colors file to keep relevent stuff jux'd.

As always, this includes some style cleanups to better vertically
align stuff as this makes things more easy to mass edit. I also
updated the typography docs to show the new explicit text/bg style
classes and document the colors in a more compact manner using
tables.

imageobzfx
imagexagmb
imagenv99e
image2tdd9

@jcredding ready for review.


jcredding (approved):

PR #85, opened 2016-12-30
apply color style classes to buttons, label, tables, grid-tables

kellyredding:
This takes all of the work to add in color style classes and
adds corresponding style classes to apply colors to the CSS
components.  This is the final step in adding standard color style
class support to Romo.

The goal here is to limit the colors available to a consistent set
and to have that set applied as you would expect throughout Romo.
This means that buttons, labels, and table backgrounds will all
match if given the same color style class.

To accomplish all this, I had to do a few extra things:

* add in button vars for the explicit colors.  These needed to be
  defined to support the button style classes in the same way the
  emphasis style classes were implemented.  I also chose to switch
  to `inverseColor` instead of hard coding `#fff` for the text/btn
  color vars.  This is all to keep colors consistent.
* I basically rewrote the docs for each CSS component.  I tried
  to remove as much cruft and superfulous markup as possible, esp
  since I was adding a significant amount of additional markup for
  documenting the color classes.

Buttons:
buttons

Labels:
labels

Tables:
tables

Grid Tables:
grid-tables

@jcredding sorry for the bloat on this PR.  The docs kinda spiraled out of control.  This docs rewrite is a preview of some optimizations I want to do for the docs in the future.  This should be the last of the Romo color stuff (for now).  Ready for review.

### DIFF
--- a/assets/css/romo/_vars.scss
+++ b/assets/css/romo/_vars.scss
@@ -129,23 +129,23 @@ $mutedBorder:      darken(adjust_hue($mutedBg, -10), 7%);
 
 $btnWarningBg:      $orange;
 $btnWarningBgHover: darken($btnWarningBg, 5%);
-$btnWarningColor:   #fff;
+$btnWarningColor:   $inverseColor;
 
 $btnDangerBg:       $red;
 $btnDangerBgHover:  darken($btnDangerBg, 10%);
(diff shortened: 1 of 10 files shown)


jcredding (approved):
@kellyredding - Really liking the docs changes, good stuff :boom:

PR #86, opened 2017-01-01
add RomoDate util functions; use in datepicker

kellyredding:
The datepicker component had some private methods for working with
dates.  However, I found I wanted to reuse the date parsing
functions in another app.  This extracts the datepicker methods
into a RomoDate namespace and generalizes them a bit for reuse.
This also adds a few extra methods for things I was doing manually
both in the datepicker and my app.

This also reworks the datepicker to take advantage of these util
methods and removes all manual date handling.  Note: as a side
effect, this switches the datepicker to always use local dates
zero'd out time.  This avoids any weird timezone issues with the
picker determining todays date.  This should have been done all
along but was missed.

Some notes on the RomoDate functions:

* all dates returned are local dates with zero time.  since these
  are date utilities, no time information is needed or should have
  any effect on the calculations.  The `.date()` method casts
  dates in this manner
* the methods are designed to be as functional as possible.  Any
  given dates will not be modified; new date objs are returned.
* the methods are designed to build on each other as much as
  possible to reuse the logic.
* there are many more utility functions that *could* be added
  (such as a "next Mon" function, etc).  I chose not to add any
  functions that I don't have a need for in the datepicker or
  my app.  As needs arise in my apps, I'll add more methods.

gif

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/date.js
+++ b/assets/js/romo/date.js
@@ -1,0 +1,310 @@
+var RomoDate = {
+
+  date: function(date) {
+    if (date === undefined || date === null) {
+      return undefined;
+    } else {
+      var d = new Date(date);
+      d.setHours(0,0,0,0); // RomoDate Dates are always local with zero time
+      return d;
+    }
+  },
+
+  for: function(yearNum, zeroBasedMonthNum, dayNum) {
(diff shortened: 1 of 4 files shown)


jcredding (approved):

PR #88, opened 2017-01-04
rename the "invoke" js component to "ajax"

kellyredding:
This better communicates intent of the component: to wrap a manual
ajax call when actions on an element should spawn and ajax call.

Not only is the component itself being renamed.  The data attrs,
event namespace, and methods are also being renamed:

* I chose to rename the event namespace to include `romo-*`.  This
  is part of the "spirit" of Romo: that it is unobtrusive.  It was
  particularly obvious on this component since "ajax" is so
  generic.  There could be other things emitting "ajax:*" events
  and I don't want Romo to interfere with those so now this
  component emits `romo-ajax:*` events.  Eventually, I'd like to
  switch all of the other components to emit similarly.
* I renamed the "load" terminology to "call".  This fits better
  now that it is named "ajax" and also it is making ajax "calls".
  This includes renaming the "invokeMethod" to "callMethod".
* I kept "invoke" around for describing things used to "kick off"
  an ajax call: invoke on/queued/running, `doInvoke`, etc.  This
  is still appropriate and was the original reason I named the
  component "invoke" in the first place: and elem that "invokes"
  an ajax call.
* I renamed any "href" references to "url".  This matches what the
  ajax call API calls it.

This is a big fat breaking change.  This will need to be rolled
out to apps carefully.  I'm choosing to add this breaking change
now to coincide with rolling out the removal of `romo-text-error`.
I prefer to limit how often breaking changes are released so I
want to release these together.

@jcredding ready for review.


kellyredding:
@jcredding this is only the component itself.  If we are good with these names, I'll update the dropdown, modal, inline and tooltip components to use/expect these names since they compose ajax.

jcredding:
@kellyredding - In your commit, first sentence, "should spawn _an_ ajax call"

jcredding (approved):
@kellyredding - I'm good with the rename :boom: Good call on using `romo-ajax` for the events.

Criptak:
@kellyredding I'm good with the rename

kellyredding:
@jcredding @Criptak thanks guys.  I've updated the js components that compose RomoAjax.  Ready for final review.  I'll test in one of our apps as well.

jcredding:
@kellyredding - Updates look good :boom:

PR #89, opened 2017-01-11
css: update border radius circle style classes

kellyredding:
This updates the circle border radius style classes to use the 
`50%` value.  This is closer to the spirit of these style classes.
If your elem is rectangular, the border will be an oval.  If it is
square, the border will be a true circle.

This previous behavior of the circle style classes produced more
of a "pill" shaped border (if rectangular, if square close to a
circle).  To preserve the old behavior and to name it more 
accurately, I've added -pill style classes.

imagesjsnr

imageuhvq3

@jcredding ready for review.

### DIFF
--- a/assets/css/romo/base.scss
+++ b/assets/css/romo/base.scss
@@ -101,7 +101,7 @@ a.romo-text {
 /* images */
 
 .romo-img-rounded { @include border-radius(6px); }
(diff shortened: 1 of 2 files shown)


kellyredding:
@jcredding OK, this is really ready for review now.

jcredding:

PR #90, opened 2017-01-14
indicator text input: update to allow pos on left or right

kellyredding:
This updates the indicator text input to allow positioning the
indicator on either the left or the right.  This is prep for
adding a currency text input which will need to place the
currency indicator on the left.

This also cleans up the indicator padding calculations to
detect the indicator elem width so it doesn't have to be manually
set unless you want to use a custom value.  I also made it so you
can customize the padding value to use to account for non-std
borders and such.

I chose to keep the indicator elem positioned on the right by
default but updated the select dropdown and datepicker to manually
set their indicator pos value in case the default changes.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/indicator_text_input.js
+++ b/assets/js/romo/indicator_text_input.js
@@ -6,8 +6,10 @@ $.fn.romoIndicatorTextInput = function() {
 
 var RomoIndicatorTextInput = function(element) {
   this.elem = $(element);
-  this.defaultIndicatorClass   = undefined;
-  this.defaultIndicatorWidthPx = 0;
+
+  this.defaultIndicatorClass     = undefined;
+  this.defaultIndicatorPaddingPx = 5;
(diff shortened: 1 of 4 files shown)


assets/js/romo/indicator_text_input.js
```diff
@@ -75,12 +77,18 @@ RomoIndicatorTextInput.prototype.doPlaceIndicatorElem = function() {
...
+
+    // add a pixel to account for the default input border
+    this.indicatorElem.css(indicatorPosition, indicatorPaddingPx+1);
+
+    // left-side padding
     // + indicator width
-    // + right-side spacing
-    var indicatorPaddingPx = 4 + indicatorWidthPx + 4;
-    this.elem.css({'padding-right': indicatorPaddingPx + 'px'});
+    // + right-side padding
```
jcredding: @kellyredding - really minor, but now that you can customize position I don't think this is always "right-side" padding

PR #91, opened 2017-01-17
add currency text input component; add RomoCurrency format functions

kellyredding:
This adds a new text input component for entering currency values.
This input is an indicator text input that can optionally format
the input on change.  It proxies all indicator text input data
attrs and events as well.

This also adds functions to a new `RomoCurrency` namespace for
formatting and un-formatting currency strings.  These are used
internally by the currency text input but they seemed handy
so I decided to make stand-alone functions out of them.

Here's an example of this in action:
gif

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/currency_text_input.js
+++ b/assets/js/romo/currency_text_input.js
@@ -1,0 +1,168 @@
+$.fn.romoCurrencyTextInput = function() {
+  return $.map(this, function(element) {
+    return new RomoCurrencyTextInput(element);
+  });
+}
+
+var RomoCurrencyTextInput = function(element) {
+  this.elem            = $(element);
+  this.hiddenInputElem = undefined;
+
+  this.defaultIndicatorPosition       = 'left';
+  this.defaultInputNameSuffix         = "romo-currency-display"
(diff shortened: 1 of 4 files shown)


assets/js/romo/currency_text_input.js
```diff
@@ -0,0 +1,168 @@
...
+RomoCurrencyTextInput.prototype._getFormatNumDecimalPlaces = function() {
+  var places = this.elem.data('romo-currency-text-input-format-num-decimal-places');
+  return !isNaN(places = Math.abs(places)) ? places : this.defaultFormatNumDecimalPlaces;
+}
+
+RomoCurrencyTextInput.prototype._getHiddenInputElem = function() {
+  return $('<input type="hidden" name="'+this.elem.attr('name')+'" value="'+this.elem.val()+'">');;
+}
+
+RomoCurrencyTextInput.prototype._getInputNameSuffix = function() {
```
jcredding: @kellyredding - You might call this `getInputName` and drop the suffix. I think this is returning the full input name.

jcredding (approved):

PR #93, opened 2017-01-19
sortable: add api to manually bind draggable elems

kellyredding:
This can be used to re-bind the sortable dragging logic to
draggable elems that have been inline edited or regenerated.
This provides a `doBindDraggableElems` method and a
`sortable:bindDraggableElems` trigger event that takes an elem
collection and re-binds all dragging logic on those elems.

This also reworkes the initial drag binding logic to use the
new method.

@jcredding ready for review.


jcredding (approved):

PR #94, opened 2017-01-24
select: position carets like indicator text inputs position indicators

kellyredding:
This updates selects to position their carets like indicator text
inputs position their indicators.  This includes:

* detect caret elem width - no need to always specify a custom
  with if using carets
* api for setting custom caret padding px - defaults to 5px like
  indicator text inputs do
* api for setting custom caret position - this allows you to set
  the caret on the left if you want but default to placing the
  caret on the right
* we can remove the custom css right pos hard-coded value - this
  was super fragile anyway - we don't need it now that we allow
  specifying padding px and left/right position

This is just to keep placing indicators like this consistent and
add some nice customizations to selects.

Here is the regular select caret handling:
norm-select

Here is a customized caret, placing it on the left width 10px of padding on each side:
odd-select

@jcredding ready for review.

### DIFF
--- a/assets/css/romo/select.scss
+++ b/assets/css/romo/select.scss
@@ -39,9 +39,10 @@
   width: 100%;
 }
 .romo-select-caret {
+  display: inline-block;
   position: absolute;
-  right: 4px + 1px; /* 4 px for desired spacing + 1 px select styling optical illusion */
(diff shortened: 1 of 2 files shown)


jcredding:

PR #95, opened 2017-02-07
Make non-word characters word boundaries for selects filter

jcredding:
This updates selects filter to use non-word characters as word
boundaries. This makes it simpler to filter options that have
non-word characters. For example, if an option had "[test]" typing
"test" into the filter previously didn't show the "[test]" option;
with this change it will.

### DIFF
--- a/assets/js/romo/select_dropdown.js
+++ b/assets/js/romo/select_dropdown.js
@@ -171,7 +171,10 @@ RomoSelectDropdown.prototype.doFilterOptionElems = function() {
     this.optionFilter.val(),
     this.romoDropdown.bodyElem.find('LI[data-romo-select-item="opt"]'),
     function(elem) {
-      return elem[0].textContent;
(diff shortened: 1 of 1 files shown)


assets/js/romo/select_dropdown.js
```diff
@@ -171,7 +171,10 @@ RomoSelectDropdown.prototype.doFilterOptionElems = function() {
     this.optionFilter.val(),
     this.romoDropdown.bodyElem.find('LI[data-romo-select-item="opt"]'),
     function(elem) {
-      return elem[0].textContent;
+      // The romo word boundary filter by default considers a space, "-" and "_" as word boundaries.
+      // We want to also consider other non-word characters (such as ":", "/", ".", "?", "=", "&")
+      // as word boundaries as well.
+      return elem[0].textContent.replace(/\W/g, ' ');;
```
kellyredding: @jcredding dem double-semicolons.

kellyredding (approved):
@jcredding one minor comment - otherwise 👎

PR #97, opened 2017-06-19
update to require a parent `.romo` elem for CSS styles to apply

kellyredding:
This allows bringing in Romo CSS alongside legacy/other CSS and
ensures Romo won't conflict with those styles.  For Romo to work
or take effect, it now has to have a parent elem with the "romo"
style class.

For most only-Romo uses, just add the "romo" class to the BODY
elem and you are good to go.  For legacy uses that want to blend
in Romo alongside some other CSS, contain your Romo markup in a
DIV (or whatever) elem with the "romo" class.

Because this causes legacy uses to break, this will need to be
deployed in a breaking change version release.

@jcredding ready for review.

### DIFF
--- a/assets/css/romo/base.scss
+++ b/assets/css/romo/base.scss
@@ -1,571 +1,580 @@
 @import 'css/romo/vars';
 @import 'css/romo/mixins';
 
-/* Globals */
-/* ------- */
+html { height: 100%; }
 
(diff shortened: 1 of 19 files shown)


jcredding (approved):
@kellyredding - This makes sense to me and is what I expected. I'm good with this :boom:

PR #98, opened 2017-06-22
add `romo-cursor-not-allowed` style class

kellyredding:
We use this cursor a bunch within the internal styles but have no
way to specify it manually with a style class.  One of the apps
we work on needed this so I am adding it to the API.

@jcredding ready for review.


jcredding (approved):

PR #99, opened 2017-06-22
add RomoAjax opt to disable the default invoke on

kellyredding:
The idea is that you should be able to use the RomoAjax component
and not have an invoke-on event.  You would do this when you just
want to invoke the component using the trigger event manually.

This prevents you from having to set some rando invoke-on value
like `data-romo-ajax-invoke-on="none"` or something.  Plus, this
turns off all invoke-on logic if no value is set.

This came up in one of our apps where we were manually triggering
the invoke on an interval.  We chose to bind RomoAjax to a visible
elem (b/c why not?) and we noticed that everytime you clicked this
random elem the ajax request would invoke which was not intended.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/ajax.js
+++ b/assets/js/romo/ajax.js
@@ -8,7 +8,12 @@ var RomoAjax = function(element) {
   this.elem       = $(element);
   this.targetElem = $(this.elem.data('romo-ajax-target'));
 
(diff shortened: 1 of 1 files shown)


jcredding (approved):

PR #100, opened 2017-06-22
add placeholder styles for romo form placeholder inputs

kellyredding:
See https://developer.mozilla.org/en-US/docs/Web/CSS/::placeholder
and https://css-tricks.com/almanac/selectors/p/placeholder/. I
wanted some custom styles to make placeholders look obviously
different from the typed in input text.  I chose to use our muted
text color and italics.

Chrome:
chrome

Firefox:
firefox

@jcredding ready for review.

### DIFF
--- a/assets/css/romo/_mixins.scss
+++ b/assets/css/romo/_mixins.scss
@@ -377,6 +377,11 @@
   color: $inputColor;
 }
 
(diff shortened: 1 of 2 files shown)


jcredding (approved):

PR #101, opened 2017-07-11
add `romo-flex-align-*` style classes

kellyredding:
This is the next chapter in the ever unfolding saga that is 
vertical alignment in CSS.  Flexbox makes it a lot easier but the
API is involved and hard to commit to memory.  These styles classes
mimic the normal alignment style classes but use flex to implement
the behavior so the results are more predictable.

Note: these should just *replace* the normal alignment classes.
Flex is heavier and should only be used when necessary.  For
typical, boring, horizontal alignment, the normal classes should
be preferred.

imagesd1si

@jcredding ready for review.


jcredding (approved):

PR #102, opened 2017-07-16
invoke RomoAjax with custom data

kellyredding:
This updates the RomoAjax component to allow invoking it with
custom data.  This can only be done with manual invocations using
the `romoAjax:triggerInvoke` event.

To use custom data when invoking, first disable the default invoke
by setting `data-romo-ajax-disable-default-invoke-on="true"` in
the component markup.  Also do not set any `romo-ajax-invoke-on`
values in the markup.  This disables any invocations that can't
accept custom data (like the default 'click' event).  Second,
in your code do something like this:

```javascript
// bind a function to the event you want to trigger the invoke on
// and in that function do ...
var customInvokeData = {'some': 'data'};
romoAjaxElem.trigger('romoAjax:triggerInvoke', customInvokeData);
```

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/ajax.js
+++ b/assets/js/romo/ajax.js
@@ -21,6 +21,8 @@ var RomoAjax = function(element) {
   this.invokeQueued  = false;
   this.invokeRunning = false;
 
+  this.queuedInvokeData = undefined;
+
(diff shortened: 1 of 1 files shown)


jcredding (approved):

PR #103, opened 2017-08-10
break out OptionListDropdown component from SelectDropdown

kellyredding:
This is a new component that contains all of the logic related to
a dropdown with a list of selectable options and a filter input
for selecting those options.  It also contains all of the dropdown
handling and navigating prev/next options.

This is prep for adding a new "picker" component.  This new
component will be very similar to the "select" component.  It too
will compose the OptionListDropdown as it will have a list of
selectable options in a dropdown.  The main difference is that
the selectable options won't be sourced from markup but instead
be fetched via RomoAjax as the user types in the filter input.
This component will be designed to work well with large sets of
options that are too cumbersome to render inline as markup.

The OptionListDropdown is designed to allow defining ad-hoc
options via the `doSetListItems` method.  It also triggers a
an event when the filter is typed that can implement custom
behavior.  The SelectDropdown listens for this event and does
its word boundary filter logic.  The picker will listen for this
event and use RomoAjax to fetch matching options and manually
rebuild the option list.

gif

@jcredding ready for review.

### DIFF
--- a/assets/css/romo/_mixins.scss
+++ b/assets/css/romo/_mixins.scss
@@ -360,7 +360,7 @@
   @include box-shadow((inset 0 1px 1px rgba(0, 0, 0, 0.075), 0 0 8px rgba(82, 168, 236, 0.6)));
 }
 
-@mixin select-optgroup {
+@mixin input-option-list-optgroup {
   font-size: $fontSize0;
   line-height: $lineHeight0;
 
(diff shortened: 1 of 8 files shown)


jcredding (approved):
@kellyredding - This makes sense and everything looks like what I expect after the discussions we've had :boom:

PR #104, opened 2017-08-11
add Picker component

kellyredding:
This adds the css and js for a new Picker component.  This is an
input that looks and behaves similarly to the Select component.
The main difference it that it has no options to choose by default
and instead goes and fetches option data as you type in the filter
input using RomoAjax.

This component is intended to look and feel just like a Select.
Therefore, the css is a copy of the Select css with the names
changed from "select" to "picker".  The js is a copy of the Select
js with a few differences.  It listens for the option list filter
change event and triggers romo ajax on filter change.  It also
listens for romo ajax success/error events and updates the list
options accordingly.  The only main difference in behavior is that
the picker has no options listed by default.

In working these changes up there were a few other loosly related
changes I did:

* cleaned up the select js to better match the organization and
  conventions in the option list js - the select js and picker js
  now match in this regard.
* also updated the indicator text input to also match org and
  style conventions above.
* updated indicator text input to make its indicator elem (the
  indicator icon) an indicator component (spinner).  This allows
  the picker to spin the indicator icon when filtering options.
  This includes proxying the trigger events from picker down to
  indicator text input.
* renamed option list dropdown's `optionFilter` to be named
  `optionFilterElem` to match our modern naming conventions.
* updated option list dropdown to always expect `displayText`
  attrs on its option data.  This is b/c the select and picker
  want text values to set on their inputs when an option is
  selected.
* added some helper methods to get selected item text/value on
  the option list dropdown.  The select and picker now use these
  to get this data as needed instead of manually deriving it or
  manually looking it up.  The picker demanded this api as it
  has no markup to lookup the selected option.  I made the select
  js match to keep them consistent.

@jcredding ready for review.

### DIFF
--- a/assets/css/romo/picker.scss
+++ b/assets/css/romo/picker.scss
@@ -1,0 +1,47 @@
+@import 'css/romo/vars';
+@import 'css/romo/mixins';
+
+.romo {
+
+  .romo-picker-wrapper {
+    display: inline-block;
+    text-align: left;
+    text-shadow: 0 1px 1px rgba(255, 255, 255, 0.75);
+    cursor: pointer;
+    @include user-select(none);
+    position: relative;
+  }
+
+  .romo-picker {
+    min-width: 50px;
+    font-weight: normal;
+    text-align: left;
+    padding-left: 6px;
+    padding-right: 6px;
+  }
+
+  .romo-picker:focus,
+  .romo-picker.romo-picker-focus { @include input-focus; }
+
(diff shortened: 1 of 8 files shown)


jcredding (approved):

kellyredding:
@jcredding cool.  I'm going to go ahead and merge this since it is kinda big.  I know there are still bugs in this implementation but I'll open a follow up PR to address those.  FYI.

PR #105, opened 2017-08-14
update picker js to properly lookup display text when setting values

kellyredding:
The picker, unlike the select, doesn't have markup to go lookup
extra meta on its options.  This especially comes into play when
dealing with showing a picker with a current value already set.
Because we didn't call to the server to get this option and it
wasn't chosen from the option list dropdown, we don't know what
display text to show in the dropdown elem for the given preset
value.

This adds `doSetSelectedValueAndText` methods to both the picker
and the option list dropdown.  This handles setting the selected
value/text on the option list dropdown element.  This is needed
b/c the picker doesn't keep its options around once the dropdown
is closed and so can't lookup selected data from option markup.

Now that we have selected data referenced in markup for both
selects and pickers, we need to handle setting a single value
on the picker.  Again, the picker doesn't have option markup to
go lookup the display text for a given value.  Instead, manually
ajax call the picker url with a `'value'` data param.  This will
signal the server to just return a list with a single result for
the item with the given value.  This single result is used to
set the selected value/text.  Again, this should only come into
play on init when setting preset values.

Finally, this updates the `'pageshow'` event handler that handles
back button usage.  In this case we can lookup the selected
value/text from the option list drodown as we are storing it on
its elem markup.

Note: the romo ajax needs to be init/bound on the picker elem,
not the option list dropdown elem.  The option list dropdown elem
is a dropdown and it already had special behavior attached to romo
ajax events due to how we use dropdowns loaded with ajax content.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/picker.js
+++ b/assets/js/romo/picker.js
@@ -26,7 +26,8 @@ var RomoPicker = function(element) {
   $(window).on("pageshow", $.proxy(function(e) {
     var selectedVal = this.romoOptionListDropdown.selectedItemValue();
     if (selectedVal !== this.elem[0].value) {
-      this.doSetValue(selectedVal);
+      var selectedText = this.romoOptionListDropdown.selectedItemText();
+      this.doSetSelectedValueAndText(selectedVal, selectedText);
     }
   }, this));
 
@@ -50,7 +51,22 @@ RomoPicker.prototype.doRefreshUI = function() {
 }
 
 RomoPicker.prototype.doSetValue = function(value) {
-  this.romoOptionListDropdown.doSetNewValue(value);
+  $.ajax({
+    type:    'GET',
+    url:     this.elem.data('romo-picker-url'),
+    data:    { 'value': value },
+    success: $.proxy(function(data, status, xhr) {
+      if (data[0] !== undefined) {
+        this.doSetSelectedValueAndText(data[0].value, data[0].displayText);
+      } else {
+        this.doSetSelectedValueAndText('', '');
+      }
+    }, this),
+  });
+}
+
+RomoPicker.prototype.doSetSelectedValueAndText = function(value, text) {
+  this.romoOptionListDropdown.doSetSelectedValueAndText(value, text);
   this._setNewValue(value);
 }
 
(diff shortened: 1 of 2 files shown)


assets/js/romo/indicator_text_input.js
```diff
@@ -68,20 +68,21 @@ RomoIndicatorTextInput.prototype._bindElem = function() {
   this.indicatorElem = $();
   var indicatorClass = this.elem.data('romo-indicator-text-input-indicator') || this.defaultIndicatorClass;
   if (indicatorClass !== undefined && indicatorClass !== 'none') {
-    this.indicatorElem = $('<i class="romo-indicator-text-input-indicator '+indicatorClass+'"></i>');
-    this.indicatorElem.romoIndicator();
+    this.indicatorElem = $('<div class="romo-indicator-text-input-indicator"><div><i class="'+indicatorClass+'"></i></div></div>');
```
jcredding: @kellyredding - Did you mean to include these changes? I don't think they were in the commit message.

assets/js/romo/picker.js
```diff
@@ -50,7 +51,18 @@ RomoPicker.prototype.doRefreshUI = function() {
 }
 
 RomoPicker.prototype.doSetValue = function(value) {
-  this.romoOptionListDropdown.doSetNewValue(value);
+  $.ajax({
+    type:    'GET',
+    url:     this.elem.data('romo-picker-url'),
+    data:    { 'value': value },
+    success: $.proxy(function(data, status, xhr) {
+      this.doSetSelectedValueAndText(data[0].value, data[0].displayText);
```
jcredding: @kellyredding - You might need to handle if `data` is empty here. It's possible we could set an invalid value and in that case it should probably not show anything selected?

jcredding (approved):

assets/js/romo/indicator_text_input.js
kellyredding: @jcredding good catch - I'm moving them to a follow up PR.

assets/js/romo/picker.js
kellyredding: @jcredding yeah good call.  I was thinking I'd just assume you'd always get something for a preset value but that probably isn't good at this component level.

PR #106, opened 2017-08-15
update indicator to calc its size based on container size

kellyredding:
The previous implementation was based on font size calculations.
This is problematic b/c the calculations assumed the input/btn/etc
had 4 pixels of top/bottom padding such that the indicator would
fill the height of the container.  However, when I went to use
this in the option list dropdown filter UI, it removes the padding
from the filter input and sets a custom input height.  This meant
the indicator greatly overflowed the input height.  This really
just exposed how weak and assumption prone the previous
implementation was.

This new implementation bases the indicator size off a given
`basisSize`.  This basis size is gathered in a few ways.  By
default, it is calculated by getting the lesser of the container's
width/height.  This ensures that by default the indicator will fit
inside its container.  However, in the case this isn't desired,
you can specify a basis size via a data attr or by passing a basis
size when you trigger the start event.  Taking a basis size on
event trigger allows for dynamic calculated sizes and also allows
you to delay calculating the size you want until start-time.

To accommodate passing in custom basis sizes, I had to switch to
building the spinner opts and spinner until start-time.

The indicator text input has also been updated to proxy the event
basis size and the basis size data attr. It has also been updated
to properly nest and structure the indicator item in a container
so that the container can be used as the indicator (spinner)
component.  This means that things are positioned properly.  The
text input's indicator elem is absolutely positioned and absolutely
positioned elems can't be indicator components b/c when the
indicator is started it sets the elem to relative position which
breaks absolute positioning.  By nesting a container and making it
the indicator, we can safely position it relatively.

Finally, this updates the option list dropdown to trigger its
indicator start event with the input height of the filter.  This
allows the indicator to properly fill the height of the filter
input.

Example of indicator in picker filter UI:
imagee8scm

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/indicator.js
+++ b/assets/js/romo/indicator.js
@@ -8,29 +8,10 @@ $.fn.romoIndicator = function() {
 }
 
 var RomoIndicator = function(element) {
-  this.elem = $(element);
-  this.spinnerOpts = {
-    lines: 11, // The number of lines to draw
-    width: 2, // The line thickness
-    length: parseInt(this.elem.css('font-size')) / 2, // The length of each line
-    radius: parseInt(this.elem.css('font-size')) / 3, // The radius of the inner circle
-    corners: 1, // Corner roundness (0..1)
-    rotate: 0, // The rotation offset
-    direction: 1, // 1: clockwise, -1: counterclockwise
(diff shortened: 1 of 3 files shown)


jcredding (approved):

PR #107, opened 2017-08-18
more small fixes to the picker and option list dropdown

kellyredding:
This comes from live testing in an app:

* fix a typo in option list dropdown that prevented typing on the
  dropdown elem to open and filter the list
* added a comment to not filter the options on close to protect
  from breaking selecting an option in the picker
* switched the picker 'value' param to 'values'.  this is prep for
  supporting a multi picker as the "value" of the input would be
  a delimeted string of many values.
* switched the picker to reset the list to default values when the
  filter value is an empty string.  this was the originally
  intended behavior but was missed.
* added missing opt list trigger when the picker sets list items.
  this was missed previously and is needed to highlight the first
  item as you are filtering/changing the items in the list.
* don't include the default options (empty val if config'd) when
  setting filtered results.  this matches how the select filters.

gif

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/option_list_dropdown.js
+++ b/assets/js/romo/option_list_dropdown.js
@@ -286,6 +286,11 @@ RomoOptionListDropdown.prototype._bindDropdownOptionFilter = function() {
   }, this));
   this.romoDropdown.elem.on('dropdown:popupClose', $.proxy(function(e, dropdown) {
     this.optionFilterElem.val('');
+    /*
+    don't call `_filterOptionElems()` here.  we need to keep the option markup as is
+    until the popup is opened again so selecting an option works.  selecting an option
+    depends on the selected item elem method which requires the markup to be in place
+    */
   }, this));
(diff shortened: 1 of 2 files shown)


assets/js/romo/option_list_dropdown.js
```diff
@@ -286,6 +286,11 @@ RomoOptionListDropdown.prototype._bindDropdownOptionFilter = function() {
   }, this));
   this.romoDropdown.elem.on('dropdown:popupClose', $.proxy(function(e, dropdown) {
     this.optionFilterElem.val('');
+    /*
+    don't call `_filterOptionElems()` here.  we need to keep the option markup as is
+    until the popup is opened again so selecting an option works.  selecting an option
+    depends on the selected item elem method which requires the markkup to be in place
```
jcredding: @kellyredding - "markkup" typo

jcredding (approved):

PR #109, opened 2017-08-22
add picker option for accepting "custom" values

kellyredding:
And by custom we mean anything typed into the filter input.  With
the option set, as you type in the filter input, the select adds
an option to the end of the list matching what you've typed.  This
is now an option that can be selected.

Note: this option's value also matches what's been typed in the
input.  Anything processing the input value needs to be aware and
handle things appropriately.

Note: I added an `optionListItems` attr to the option list 
dropdown thinking I would need it but ended up not using it on 
this.  I chose to leave it in though b/c it doesn't hurt.  I also
proxied it through the select dropdown so both the picker and the
select can use it if a need comes up.

with prompt:
gif

without prompt:
wo-prompt

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/option_list_dropdown.js
+++ b/assets/js/romo/option_list_dropdown.js
@@ -5,8 +5,10 @@ $.fn.romoOptionListDropdown = function() {
 }
 
 var RomoOptionListDropdown = function(element) {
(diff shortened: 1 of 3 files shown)


jcredding (approved):

PR #110, opened 2017-08-23
rework how select dropdowns build their option list dropdown items

kellyredding:
This switches to doing all operations on the options markup elems
and then rebuilding the option list dropdown items from the
markdown.  This is prep for adding an option for to allow custom
values (like the picker has).  In order to do that feature, I need
selects to rebuild their option list as they are filtered.

The filter is now setup to filter the option elems, not the option
list dropdown item elems.  These option elems are marked with the
hidden class on filter.  The code to rebuild the option list
dropdown items ignores any hidden options.

Note: this means that nothing is using the option list dropdown's
`data-romo-option-list-dropdown-item-selector-customization` attr
anymore.  I chose to keep it in for now in case a need arises in
the future - the feature is still valid if not needed at this
moment.

Note: this also fixes yet another select prev/next bug.  I noticed
this when testing on a select with many groups.  The prev/next
logic would fail if the next/prev list had no options.  This
updates the logic to keep trying next/prev elems until it finds
an optgroup list with options, an option, or nothing.

Grouped example:
grouped

Ungrouped example:
ungrouped

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/option_list_dropdown.js
+++ b/assets/js/romo/option_list_dropdown.js
@@ -504,6 +504,11 @@ RomoOptionListDropdown.prototype._nextListItem = function() {
     curr = curr.closest('UL.romo-option-list-optgroup') || curr;
     next = Romo.selectNext(curr, this.itemSelector+', UL.romo-option-list-optgroup');
   }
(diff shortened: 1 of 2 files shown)


jcredding (approved):

PR #111, opened 2017-08-23
update select dropdowns to not render optgroups without child opts

kellyredding:
This keeps things clean when filtering grouped selects.  There is
no need to show a bunch of empty group lists as a user types and
filters down the option list.  This is possible b/c in PR 110 we
switched to re-rendering the list on filter changes.

Note: a side-effect of this is that any groups without options
at all in the markup will also not be displayed.

See #110 for reference.

Example:
filter-out

@jcredding ready for review.


jcredding (approved):
@kellyredding - Looks good :boom: This is definitely much cleaner when filtering the options.

PR #112, opened 2017-08-23
add select dropdown option for accepting "custom" values

kellyredding:
This is similar to the same effort done for pickers in PR 109. By
"custom" we mean anything typed into the filter input.  With the
option set, as you type in the filter input, the select adds an
option to the end of the list matching what you've typed.  This is
now an option that can be selected.

Note: this option's value also matches what's been typed in the
input.  Anything processing the input value needs to be aware and
handle things appropriately.

Note: I updated the select component to proxy its data attrs to
the select dropdown so selects get the same behavior.

See #109 for reference.

cust-select-opt

@jcredding ready for review.


jcredding (approved):

PR #113, opened 2017-08-23
add select/picker/option-list-dropdown option to open on focus

kellyredding:
This option will open the popup and focus the filter input when
the dropdown elem is focused.  This is a UX convenience that can
optionally be turned on.  It is particularly helpful for single
input forms that submit on input change.  No need to make the user
click to launch the form then click to open the popup then start
typing, etc.

The data attr and logic live on the option list dropdown: the
select dropdown, select, and picker components proxy the option
so all of these components get the same behavior.

Select:
open

Picker:
picker-open

@jcredding ready for review.


jcredding (approved):
@kellyredding - Nice feature, I'm good with this :boom:

PR #114, opened 2017-08-24
update datepicker, select, and picker "itemSelected" events

kellyredding:
These events should never have been publishing the new/prev values
in their payloads - that is what the "change" event is for.  Also,
the previously didn't distinguish from any item being selected vs
just a new item being selected.

This updates the event payloads to just include the selected item
value.  This also adds publishing a second "newItemSelected" event
that only fires when a not previously selected item is selected.
Note: for the selects/pickers, the payload also includes the item
display text.

This is prep for reworking the selects and pickers to get them
ready for multi usage.  I chose to go ahead and update the
datepicker to keep it consistent with the select/picker.

Note: this also adds a helper comment to the opt list dropdown and
datepicker noting that the item selected events should always be
published before the change events.  This is by design and is
relied upon by other components so this reminds us not to change
the event firing order.  This order will also be needed when we
add the multi select/picker logic.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/datepicker.js
+++ b/assets/js/romo/datepicker.js
@@ -178,7 +178,11 @@ RomoDatepicker.prototype.doSelectHighlightedItem = function() {
   this.romoDropdown.doPopupClose();
   this.doSetDate(newValue);
   this.elem.focus();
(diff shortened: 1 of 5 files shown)


jcredding (approved):

PR #116, opened 2017-08-24
update select js refresh ui logic to match the picker js

kellyredding:
This is all prep for adding multi select/picker support.  This
is really just to keep the picker/select implementaions as
similar as possible.  Like in PR 115:

* I made the `_refreshUI` method private b/c there is no need
  to call it publicly
* I moved the refresh UI logic out of the `_setValue` method b/c I
  felt it does more than the method name implies.

I also renamed the picker's `doSetSelectedValueAndText` method to
remove "Selected" from the name.  This method is more accurate
about what it does and is more consistent with the `doSetValue`
method above and the `doSetValue` method on selects.

@jcredding ready for review.


jcredding (approved):

PR #117, opened 2017-08-25
update selects to change their option DOM to set values

kellyredding:
This not only helps the selects render properly when the back
button is used, it is also a more robust way of setting the value
when there are multiple options to select.  This is prep for
adding multi-select support in a future effort.

Note: I had to use both `attr`/`removeAttr` and `prop`.  The attr
methods change the DOM - the prop method changes the value of the
select.  I would have just used the prop method but I wanted to
change the DOM for back button support.

Note: I also moved the `_onCaretClick` function in both the
picker and select - this was just to jux it next to the other
caret methods and should have been done previously.

Note: I removed calling `_refreshUI` from the `_setValue` method
in the select js.  This should have been done in PR 116 but was
missed and resulted in `_refreshUI` being called multiple times
in succession.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/picker.js
+++ b/assets/js/romo/picker.js
@@ -279,13 +279,6 @@ RomoPicker.prototype._buildCustomOptionItem = function(value) {
   };
 }
 
-RomoPicker.prototype._onCaretClick = function(e) {
-  if (this.elem.prop('disabled') === false) {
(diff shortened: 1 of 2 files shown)


jcredding (approved):

PR #118, opened 2017-08-25
update select dropdown to sanitize its options

kellyredding:
Specifically this updates any options without a value to have an
explicit `value=""`.  The seclect and select dropdown both expect
all options to have a value.  This enforces it.

This is particularly helpful b/c we sometimes define the "default"
or "empty" option without a value b/c the "" is implied.  This
keeps things from breaking when we forget the `value=""`.

@jcredding ready for review.


jcredding (approved):

PR #119, opened 2017-08-25
fix preset value handling in picker/select

kellyredding:
This switches to more robustly handling preset values in both
the select and picker.

This fixes a bug in the selects where an "empty option" with
non-empty display text wouldn't show up if the select had no
value.  We weren't setting the empty value if the value was empty.
This changes to always setting the value on init, even if it is
empty.

This also updates the picker to match setting its preset value
on init.  However, pickers have to handle their "empty options"
differently as there aren't options in the DOM for pickers.
Instead, if the preset value is empty, we look for the empty
option setting and use a new empty option display text setting if
available.  Otherwise we just set empty value/text.  The
`doSetValue` ajax success callback is updated similarly.  Finally
we use this new empty option display text setting when building
the empty option.

Overall, this gets selects and pickers both setting their preset
values (empty or not) consistently and also provides a way for
pickers to have empty options with custom display text like
selects do.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/picker.js
+++ b/assets/js/romo/picker.js
@@ -17,6 +17,15 @@ var RomoPicker = function(element) {
   this.doInit();
   this._bindElem();
 
+  var presetVal = this.elem[0].value;
+  if (presetVal !== '') {
+    this.doSetValue(presetVal);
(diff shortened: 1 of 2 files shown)


jcredding (approved):

PR #120, opened 2017-08-25
fix custom options on selects: add custom option markup

kellyredding:
This allows selects to actually set a custom value.  Before the
dropdown would say a value was selected that there was no option
for so the value wouldn't take.  This updates the select
dropdown to actually create a custom option with the correct
value/text so it can be set as the selects value.

If a different custom option is subsequently chosen, the first
custom option is overridden.  If a non cusotm option is selected,
any custom option is removed as it is no longer needed.

gif

@jcredding ready for review.


jcredding (approved):

PR #121, opened 2017-08-29
add SelectedOptionsList js component; design picker/select interactions

kellyredding:
This is prep for adding multi support for romo pickers/selects.
This component is the common UI handling for the selected options
for each.

This component's api includes methods to set items, clear items,
add item, remove item, and refresh ui.  The design is that you
modify the item set using the api methods and then call refresh
ui to update the markup.

In addtion, each selected item's markup supports an optional
rm icon.  If clicked, the component publishes an rm icon clicked
event that can be bound to for running remove logic.  It is
expected that the rm api method would be called at some point
after the event is published.

Here are some of the other smaller UI features:

* the selected ui is synced on refresh - the items are compared to
  the ui and only the items that need to be added are added and
  the ui items that need removing are removed.  This avoids any
  ui flashes with a clear and rebuild approach.
* the lists height grows as needed up to an option max rows.  If
  no max-rows is configured, the height grows infinitely.
* the component is built with a "focus elem".  This component is
  not designed to be directly focused (it will be paired with a
  select/picker that will be focused).  If the elem is clicked, it
  focuses its focus elem.  If its focus elem is focused/blurred,
  a focus style class is added/removed to the elem to make it
  appear as it is focused/blurred as part of the focus elem being
  focused/blurred.

This is the first pass at this component and its usage. In a
coming, I'll intergrate this with selects/pickers with the multi
prop and polish the UX.

Note: I went ahead and psuedo-code added notes for integrating
with the select/picker.  I wanted to be sure I could use this
as is and integrate it as designed.  These notes will be removed
in a coming effort.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/selected_options_list.js
+++ b/assets/js/romo/selected_options_list.js
@@ -1,0 +1,185 @@
+var RomoSelectedOptionsList = function(focusElem) {
+  this.elem      = undefined;
+  this.focusElem = focusElem; // picker/select dropdown elem
+
+  this.items                  = [];
+  this.focusStyleClass        = this.elem.data('romo-option-list-focus-style-class');
+  this.defaultRmIconClass     = undefined;
+  this.defaultRmIconPaddingPx = 5;
+  this.defaultRmIconPosition  = 'right'
+  this.focusBlurTimeoutId     = undefined;
+
+  this.doInit();
+  this._bindElem();
+}
+
+RomoSelectedOptionsList.prototype.doInit = function() {
+  // override as needed
+}
+
+/*
+Options are specified as a list of items.  Each 'option' item object
+has display text and a value. They are compatible with the option list
+dropdown option items.
+
+Example:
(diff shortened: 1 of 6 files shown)


assets/js/romo/selected_options_list.js
```diff
@@ -0,0 +1,185 @@
...
+  });
+  rmNodes.each(function(rmNode) {
+    $(rmNode).remove();
+  });
+  addItems.each(function(addItem) {
+    uiListElem.append(this._buildItemElem(addItem));
+  });
+
+  var firstItemNode    = uiListElem.find('.romo-selected-option-list-item')[0];
+  var itemHeight       = parseInt(Romo.getComputedStyle(firstItemNode, "height"), 25);
```
jcredding: @kellyredding - I'm probably missing something but why are you using a radix of 25 with `parseInt`?

jcredding (approved):
@kellyredding - One comment but I'm good with this as a starting point :boom: Everything looks like what you explained and makes sense to me.

PR #123, opened 2017-08-30
selected options list items UI/UX

kellyredding:
This gets the UI and UX for the selected items in the selected
options list working.  This includes a bunch of tweaks that became
necessary when testing in the browser.

I chose to remove the "rm icon" design and instead switch to
striking thru the text on hover to signify that you can click and
remove the item.  I didn't want to clutter up the UI by adding in
a bunch of rm icon noise and the strikethru seems to get a similar
effect.  This also simplified the implementation by removing a
bunch of crappy rm icon handling code.

I also switched to calculating the max-width for an item until
after it was added to the DOM.  To properly calc this, you need
to take into account all the padding, borders, etc and not all of
that info is ready until it is in the DOM.

I removed auto refreshing the UI from the item set API methods.
This should not have been added as the design has always been to
just update the item set and let the user determine when to
refresh the UI (as part of a large refresh perhaps).

Finally, this updates some select logic to allow adding items
when they are selected.  This is just the bare minimum to test the
list UI and is incomplete.

gif

@jcredding ready for review.

### DIFF
--- a/assets/css/romo/forms.scss
+++ b/assets/css/romo/forms.scss
@@ -369,17 +369,22 @@
   /* selected options list (used by multi selects/pickers) */
 
   .romo-selected-options-list {
-    cursor: pointer;
     font-weight: normal;
-    padding: 4px 0;
     @include user-select(none);
-
     background-color: $inputBgColor;
     color: $inputColor;
     @include border1;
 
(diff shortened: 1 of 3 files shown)


jcredding (approved):

PR #124, opened 2017-08-30
get selected options list max rows handling working

kellyredding:
This updates the max rows handling to make it calculate the max
height properly and also updated it to scroll to the last item
when the max-rows height limit is needed.  This means you will
always see the item you just selected at the bottom of the list.

Note: I updated the max height calc to properly account for list
padding and item margin/borders.  I chose to also add half a row
of extra height.  This shows that there is more above or below
the list to scroll to.  Without it, it might look like there are
only what you can see.

gif

@jcredding ready for review.


jcredding (approved):

PR #126, opened 2017-08-31
integrate SelectedOptionsList with Select component

kellyredding:
This gets the custom multi select component working and fully
integrated with the selected options list UI.  I made a few
tweaks to how selects handle their values to support this:

* There are 3 main places where the integration occurs: setting
  preset/ad-hoc values in `doSetValue`, removing a value on
  `romoSelectedOptionsList:itemClick`, and adding a value on
  `selectDropdown:newItemSelected`.
* In general, I tried to do the minimum operations wherever
  possible.  This means only adding it the value isn't already
  added, the inverse for removal, and both for setting a set
  of values.
* Because we need to support multi-selects, we can no longer use
  `this.elem[0].value` to lookup the value for a select.  I wrote
  `_elemValues` to replace this.  It looks up the options that are
  selected and returns their values as an array.
* To simplify the handling between normal/multi selects, I
  switched all value handling to be array based.  Normal selects
  now just use single item value arrays while multi selects can
  have multi item value arrays.
* The only place this is not true is the public `doSetValue`
  method and corresponding trigger event.  Because this value can
  come from end-users, I check the value and cast it to an array
  if necessary.
* I also updated the `doSetValue` method to filter out values
  where there was no corresponding option.  This helps prevents
  weird edge cases.
* I rewrote the `_setValue` method to be `_setValues` and made it
  array based.  It compares the current values to the new values
  and only adds/removes options as needed.

Finally, this fixes a bug in refreshing the normal select's UI:
there are cases where no option is selected or there are no
at all.  In that case we treat the first option as selected and
return its single value with `_elemValues`.  If there are no
options, `_elemValues` returns `[""]` for normal selects and `[]`
for multi selects.

gif

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/select.js
+++ b/assets/js/romo/select.js
@@ -14,7 +14,7 @@ var RomoSelect = function(element) {
   this.doInit();
   this._bindElem();
 
-  this.doSetValue(this.elem[0].value);
+  this.doSetValue(this._elemValues());
 
   if (this.elem.attr('id') !== undefined) {
     $('label[for="'+this.elem.attr('id')+'"]').on('click', $.proxy(function(e) {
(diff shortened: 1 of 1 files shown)


jcredding (approved):

PR #128, opened 2017-09-01
update opt list dropdown focus logic to better handle open on focus

kellyredding:
This is more cleanup that came from working on the multi picker.
I was testing the multi picker with open on focus and found a
weird open toggle issue.  This came up when I clicked on the
selected options list.  The desired behavior is that it would
close the popup and just focus the opt list dropdown.  However,
b/c of the open on focus option, it close the popup, focus the
opt list dropdown and then re-open the popup.

I added a `doFocus` method on the opt list dropdown that I'm now
using to focuse the dropdown.  It takes an optional arg to specify
whether is should open on focus or not.  If given true or false
it sets that on the dropdown and then focuses it.  If no arg is
given, it leaves the open on focus attr as is.  This allows me
to declare that the popup should not open on focus when clicking
the selected options list (ie removing selected options).

I also updated the selected options list to now trigger an event
when it is clicked instead of just blindly focusing the focus
elem.  This allows me a hook to call `doFocus` appropriately on
the select and in the future on the picker.

@jcredding FYI again.  I want to go ahead and merge this so I can get the multi picker PR up and you are out today.  Check me on this, make any comments, and I'll address them in some follow up.  Thanks man.

### DIFF
--- a/assets/js/romo/option_list_dropdown.js
+++ b/assets/js/romo/option_list_dropdown.js
@@ -93,6 +93,15 @@ RomoOptionListDropdown.prototype.doSetSelectedValueAndText = function(value, tex
   this.prevValue = value;
 }
 
+RomoOptionListDropdown.prototype.doFocus = function(openOnFocus) {
+  if (openOnFocus === true) {
+    this.openOnFocus = true;
+  } else if (openOnFocus === false) {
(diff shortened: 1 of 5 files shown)


jcredding (commented):

PR #129, opened 2017-09-01
add multi picker support

kellyredding:
This updates the picker, similarly to how the select was updated
in PR 126, to support a `multiple` property for picking multiple
values.

As with selects, this integrates the selected options list with
the picker in the 3 main places: setting preset values, setting
newly selected individual items, and removing individual items.

The main difference from the select implementation is getting and
setting the values.  There isn't options markup to updated so
instead it serializes the values with a customizable delim.  To
get the values, split the elem value.  To set them, join the
values and set as the elems value.  Whether setting values in a
multi picker or not, we only ever need to set a single display
text for the dropdown.  If multi, we always set `''`, otherwise
we set the single display text.

See #126 for reference.

gif

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/picker.js
+++ b/assets/js/romo/picker.js
@@ -10,6 +10,7 @@ var RomoPicker = function(element) {
   this.defaultCaretClass     = undefined;
   this.defaultCaretPaddingPx = 5;
   this.defaultCaretPosition  = 'right'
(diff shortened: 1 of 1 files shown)


jcredding (approved):
@kellyredding - Looks like what I expect :boom: Good stuff.

PR #131, opened 2017-09-25
add in base methods to emulate jQuery's api

kellyredding:
This is prep for eventually moving to vanilla js for Romo.  These
methods represent the bulk of jQuery's api that romo uses.  This
will allow us to slowly transition off of jQuery specific logic
and onto Romo/vanilla js logic.

@jcredding ready for review.


jcredding (approved):

PR #133, opened 2017-09-26
add a hack for getting event handling working with jQuery

kellyredding:
So in testing the jQuery replacement api, it looks like jQuery 
events aren't compatible with the CustomEvents api.  I'm not
sure whether this is the CustomEvents api or jQuery's magic
and layers they put into the events they trigger.  Whatever.
See https://bugs.jquery.com/ticket/11047 for an example of 
one problem.  Ugh.

So, to get by while we transition Romo off of jQuery, this hack
changes the `Romo.on`, `Romo.off`, and `Romo.trigger`
implementations to instead use the jQuery equivalents.  Once Romo
has been rewritten to no longer need/support jQuery, this hack
will be removed.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -440,41 +440,50 @@ Romo.prototype.ajax = function(settings) {
 // events
 
 Romo.prototype.on = function(elem, eventName, fn) {
-  var proxyFn = function(e) {
-    var result = fn.apply(elem, e.detail === undefined ? [e] : [e].concat(e.detail));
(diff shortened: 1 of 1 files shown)


jcredding (approved):

kellyredding:
@jcredding please look again - I've had to expand/redo this.

jcredding (approved):
@kellyredding - Yeah I'm good with the changes on this :boom:

PR #134, opened 2017-09-26
fix a bug hitting backpace on selects/pickers

kellyredding:
Hitting backspace would put the string "Backspace" in the filter
for option list dropdowns (selects/pickers).  This updates to
check for backspace and doesn't add its key to the input but does
open the popup.

@jcredding ready for review.


jcredding (approved):

PR #136, opened 2017-09-28
Fix `Romo.ajax` GET requests with `data`, add to query string

jcredding:
This fixes an issue with the `Romo.ajax` helper making GET
requests with the `data` key set. This is expected to send the
`data` key-values as params to the server but they were being
ignored because they were being added to the body of the request.
Most web servers will ignore the body of GET requests and only
parse the query string for request params. This made it seem like
the `Romo.ajax` was not sending the `data` key-values to the
server.

To fix this, the `Romo.ajax` helper now checks if the request is
a GET request and if it has data key-values. If so, it adds the
key-values to the query string and unsets the data so it's not
passed in the request body. This allows the data key-values to
properly be seen as request params on the server.

Note: This also updates the defaults for the url to use the
current url (`window.location`) if it's not provided. This ensures
we always have a string to append to.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -433,8 +433,19 @@ Romo.prototype.decodeParamMap = [
 // AJAX
 
 Romo.prototype.ajax = function(settings) {
+  var httpMethod = (settings.type || 'GET').toUpperCase();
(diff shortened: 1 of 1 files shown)


kellyredding (approved):
@jcredding this looks good man 👎

PR #138, opened 2017-10-04
romoAjax: fix bug with invokes with data

kellyredding:
This adds a second event handler, `onTriggerInvoke` that is now
used exculsively with the 'romoAjax:triggerInvoke' event.  This
allows having separate event processing for triggered invokes
vs auto invokes (from a click for example).  Auto invokes never
expect/take data where trigger invokes do.  This fixes a bug
where you use an event that passes unrelated data as an auto
invoke.  Without this, the unrelated data would be interpreted
as invoke data which is not intended.

This also adds `return false` directives to the picker's romo
ajax event handlers.  Since romo ajax components are prone to
being nested, it is a good practice to stop propagation on any
romo ajax events.  This handles that for pickers.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/ajax.js
+++ b/assets/js/romo/ajax.js
@@ -41,19 +41,29 @@ RomoAjax.prototype.doBindElem = function() {
   if (this.invokeOn !== undefined) {
     this.elem.on(this.invokeOn, $.proxy(this.onInvoke, this));
   }
-  this.elem.on('romoAjax:triggerInvoke', $.proxy(this.onInvoke, this));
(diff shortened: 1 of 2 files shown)


kellyredding:
@jcredding I'm going to go ahead and merge this, FYI.  Please do look at it when you get a chance but I need to move on with this change.  Thanks.

jcredding:
@kellyredding - I'm good with this :boom: Sorry meant to look at it this morning.

kellyredding:
@jcredding no worries, thanks man.

PR #139, opened 2017-10-12
add base js `Romo.is` method

kellyredding:
This is part of prepping to switch off of jQuery and on to Romo's
vanilla js API.  This `is` method emulates jQuery's `is` method
with a polyfill for older browsers.  I pulled this implementation
from: http://youmightnotneedjquery.com/#matches_selector.

@jcredding ready for review.


jcredding:

PR #140, opened 2017-10-12
Add `Romo.assign` util, use to merge javascript objects

jcredding:
This adds a `Romo.assign` util that can be used to merge
javascript objects. This is similar to jquery's `extend` only
it uses the `Object.assign` function if available or it uses
a polyfill if the `Object.assign` function is not available. I
pulled the polyfill implementation from:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

This is part of an effort to make Romo no longer dependent on jquery.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -575,6 +575,31 @@ Romo.prototype.array = function(collection) {
   return Array.prototype.slice.call(collection);
 }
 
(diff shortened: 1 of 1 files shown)


kellyredding:
@jcredding this looks legit to me - thx! 👎

PR #141, opened 2017-10-12
add a base js `parents` method

kellyredding:
This is prep for moving off of jQuery and onto Romo's vanilla js
API.  This `parents` method emulatest jQuery's `parents` method,
getting a list of parent elems for a given child elem.  It also
optionally takes a selector string to filter the parents with.

Note: the diff is a little rough.  I moved the `closest` method
definition to jux it after the `parents` method since they both
look at parent elems and both take a selector.  This also jux's
the `prev|next` meths with the `select{Prev|Next}` meths.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -42,18 +42,25 @@ Romo.prototype.parent = function(childElem) {
   return childElem.parentNode;
 }
 
-Romo.prototype.siblings = function(elem) {
-  return Array.prototype.filter.call(elem.parentNode.children, function(childElem) {
-    return childElem !== elem;
-  });
-}
-
-Romo.prototype.prev = function(fromElem) {
-  return fromElem.previousElementSibling;
-}
-
-Romo.prototype.next = function(fromElem) {
-  return fromElem.nextElementSibling;
+Romo.prototype.parents = function(childElem, selector) {
+  var parentElem = this.parent(childElem);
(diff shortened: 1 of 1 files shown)


assets/js/romo/base.js
```diff
@@ -42,18 +42,25 @@ Romo.prototype.parent = function(childElem) {
...
+  if (parentElem) {
+    if (!selector || Romo.is(parentElem, selector)) {
+      if (Romo.is(parentElem, 'body')) {
+        return [parentElem];
+      } else {
+        return [parentElem].concat(this.parents(parentElem, selector));
+      }
+    } else {
+      if (Romo.is(parentElem, 'body')) {
+        return [];
```
jcredding, line 56: @kellyredding - It seems inconsistent to not return the `<body>` elem here but with the selector it can? I don't know of any issues related to this but it seems more consistent to me that the `parents` method either does or doesn't return the `<body>` elem regardless of a selector being passed.

jcredding (approved):
@kellyredding - Once comment about whether the `<body>` elem should be returned or not. I'm good with this :boom:

assets/js/romo/base.js
jcredding: @kellyredding - Thanks for explaining, I didn't interpret this correctly. I'm good with this.

PR #142, opened 2017-10-12
rework the parent-child elems handler to use Romo vanilla js

kellyredding:
This is part of reworking Romo to not require jQuery.  This
switches the parent-child elems tracking/handling to use Romo
vanilla js.  There are no behavior changes here - just switches to
the new syntax.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -701,49 +701,50 @@ Romo.prototype._addEventCallback = function(name, callback) {
   this._eventCallbacks.push({ eventName: name, callback:  callback });
 }
 
-// RomoParentChildElems (TODO: rework w/o jQuery)
+// RomoParentChildElems
 
 var RomoParentChildElems = function() {
   this.attrName = 'romo-parent-elem-id';
   this.elemId   = 0;
   this.elems    = {};
 
-  var parentRemovedObserver = new MutationObserver($.proxy(function(mutationRecords) {
-    mutationRecords.forEach($.proxy(function(mutationRecord) {
+  var parentRemovedObserver = new MutationObserver(Romo.proxy(function(mutationRecords) {
+    mutationRecords.forEach(Romo.proxy(function(mutationRecord) {
       if (mutationRecord.type === 'childList' && mutationRecord.removedNodes.length > 0) {
-        $.each($(mutationRecord.removedNodes), $.proxy(function(idx, node) {
-          this.remove($(node));
+        mutationRecord.removedNodes.forEach(Romo.proxy(function(removedNodeElem) {
+          this.remove(removedNodeElem);
         }, this));
       }
     }, this));
   }, this));
 
-  parentRemovedObserver.observe($('body')[0], {
+  parentRemovedObserver.observe(Romo.f('body')[0], {
     childList: true,
     subtree:   true
   });
 }
 
 RomoParentChildElems.prototype.add = function(parentElem, childElems) {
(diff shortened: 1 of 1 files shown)


assets/js/romo/base.js
```diff
@@ -671,49 +671,50 @@ Romo.prototype._addEventCallback = function(name, callback) {
...
-  parentRemovedObserver.observe($('body')[0], {
+  parentRemovedObserver.observe(Romo.f('body')[0], {
     childList: true,
     subtree:   true
   });
 }
 
 RomoParentChildElems.prototype.add = function(parentElem, childElems) {
-  parentElem.attr('data-'+this.attrName, this._push(childElems, parentElem.data(this.attrName)));
+  Romo.attr(parentElem, 'data-'+this.attrName, this._push(childElems, Romo.data(parentElem, this.attrName)));
```
jcredding: @kellyredding - Should this be `setAttr`? I think you are passing 3 args so the last one is intended to set the data attribute?

assets/js/romo/base.js
```diff
@@ -671,49 +671,50 @@ Romo.prototype._addEventCallback = function(name, callback) {
...
 }
 
 RomoParentChildElems.prototype.remove = function(nodeElem) {
-  if (nodeElem.data('romo-parent-removed-observer-disabled') !== true) {
-    if (nodeElem.data(this.attrName) !== undefined) {
-      this._removeChildElems(nodeElem);  // node is a parent elem itself
+  if (Romo.data(nodeElem, 'romo-parent-removed-observer-disabled') !== true) {
+    if (Romo.data(nodeElem, this.attrName) !== undefined) {
```
jcredding, line 733: @kellyredding - I think the `Romo.data` returns `null` when the data attribute isn't on the element (that's what I see when I test it in the console).

I saw these in the components and meant to ask if we wanted to change the `Romo.data` to return `undefined` instead but forgot about it until I saw this change.

jcredding (approved):
@kellyredding - A couple comments but I'm good with this :boom:

assets/js/romo/base.js
kellyredding: @jcredding I addressed this in https://github.com/redding/romo/pull/143, FYI

assets/js/romo/base.js
```diff
@@ -671,49 +671,50 @@ Romo.prototype._addEventCallback = function(name, callback) {
...
-  parentRemovedObserver.observe($('body')[0], {
+  parentRemovedObserver.observe(Romo.f('body')[0], {
     childList: true,
     subtree:   true
   });
 }
 
 RomoParentChildElems.prototype.add = function(parentElem, childElems) {
-  parentElem.attr('data-'+this.attrName, this._push(childElems, parentElem.data(this.attrName)));
+  Romo.setData(parentElem, this.attrName, this._push(childElems, Romo.data(parentElem, this.attrName)));
```
kellyredding, line 728: @jcredding I changed this to use `setData`, FYI.

assets/js/romo/base.js
```diff
@@ -724,7 +725,8 @@ RomoParentChildElems.prototype._push = function(items, id) {
   if (this.elems[id] === undefined) {
     this.elems[id] = []
   }
-  items.forEach($.proxy(function(item){ this.elems[id].push(item) }, this));
+  this.elems[id] = this.elems[id].concat(items);
+
```
kellyredding, line 759: @jcredding I changed this to use `concat`, FYI.  I should have used this originally but didn't know about it at the time.

PR #143, opened 2017-10-12
update how get/set attr and get/set data work for consistency

kellyredding:
This does a few tweaks to these methods to ensure they return
consistent values:

* We prefer `undefined` values over `null`, so the `getAttr`
  now manually checks for `null` return values and returns
  `undefined` instead.
* any `null` values are deserialized to `undefined`
* the set attr/data methods now both force the values to be
  strings and return the string values they set.
* the method checks were removed from get/set attr as there is
  good browser support for both.

There is now only one case where you can get null values out of
these methods: you use `setData` to write a `null` value.  This
gets serialized as `"null"` in `setAttr`.  Then you use `data`
to get that same attr out and `"null"` gets deserialized to
`null`.  Otherwise, any "null-ish" values will be returned as
`undefined`.

This is all part of tweaking Romo's new js API to be as
predictable and consistent as possible.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -127,14 +127,18 @@ Romo.prototype.selectPrev = function(elem, selector) {
 // attributes, styles, classes
 
 Romo.prototype.attr = function(elem, attrName) {
-  return elem.getAttribute ? elem.getAttribute(attrName) : undefined;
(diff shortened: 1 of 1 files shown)


jcredding (approved):

PR #144, opened 2017-10-12
redesign/rework the init UI elems system

kellyredding:
This is part of moving off of jQuery and on to vanilla Romo js.
This switches all of the base js methods related to initializing
markup components/elems.  This also updates the components to work
in the new system with the init callback functions they add.

There are a number of changes here:

* the init system is no longer based on events bubbling to the
  body elem.
* instead, `triggerInitUI` calls each init ui callback on each
  elem given.  It calls each callback in a `setTimeout` to kick
  the reactor to mimic how the events would invoke the callbacks.
* I had to stop using events b/c Romo custom events don't bubble
  so I couldn't count on the body element catching all init ui
  events.
* `onInitUI` still adds init callback methods
* `initUIElems` still finds all child elems of a given elem that
  match a given selector (including the given elem itself).
* updated the `init*` methods to match the new DOM manipulation
  methods: replace, update, prepend, append, before, and after -
  with the "Html" variants.
* the `init*` methods now return the elems that are manipulated.

All of this gets the component auto init system working without
jQuery or events.

@jcredding ready for review.


jcredding (approved):

PR #145, opened 2017-10-12
update base prev/next methods to take an optional selector

kellyredding:
This is part of reworking Romo to not depend on jQuery and instead
use vanilla js.  This deprecates the `selectPrev` and `selectNext`
meths and instead updates the `prev` and `next` meths to take an
optional selector.  This provides the functionality of both
and matches taking an optional selector like `parents` and
`closest` does.

@jcredding ready for review.


jcredding (approved):

PR #146, opened 2017-10-12
update base param method to use vanilla js

kellyredding:
This is part of getting off of jQuery and on to vanilla js. This
reimplements a subset of what jQuery's `param` method does.  I
chose to limit it to just array handling 1-level deep.  jQuery
does some complex stuff with serializing nested objects that I
didn't get into b/c we rarely need that level of complexity in
our apps.  We can always expand out the logic later if needed.

@jcredding ready for review.


jcredding (approved):

PR #147, opened 2017-10-12
update base js parse z-index logic to use vanilla js

kellyredding:
This is part of moving off of jquery and onto vanilla js.  This
covers the parse z-index logic.  I updated the implementation
to return the first non-zero z-index value - there was no need to
collect all z-index values and then just get the first non-zero
one.

@jcredding ready for review.  Check my implementation here.  Make sure it checks out - I haven't gotten to test it well yet.  Thanks.


jcredding (approved):
@kellyredding - This makes sense to me, good stuff :boom:

PR #148, opened 2017-10-13
Update `RomoAjax` component to not use jquery

jcredding:
This updates the `RomoAjax` component to not use jquery. This is
part of the effort to no longer require jquery to use Romo. This
removes all of the jquery usage within the `RomoAjax` component
and also updates the dropdown, modal, picker, and tooltip
components to not use the jquery method for initializing a romo
ajax component.

This removes the romo ajax component unbinding all invoke on
events from its elem when initialized. The `Romo.off` doesn't
support unbinding all events but we also decided that Romo doesn't
need to be this aggressive with its components elems. This allows
users to bind event handlers to the romo ajax's `invokeOn` event
if desired (though it's not recommended).

Note: This didn't update any event names because they are already
prefixed with `romoAjax`.

### DIFF
--- a/assets/js/romo/ajax.js
+++ b/assets/js/romo/ajax.js
@@ -1,25 +1,22 @@
 var RomoAjax = function(element) {
-  this.elem       = $(element);
-  this.targetElem = $(this.elem.data('romo-ajax-target'));
+  this.elem       = element;
+  this.targetElem = Romo.f(Romo.data(this.elem, 'romo-ajax-target'))[0];
 
   this.defaultInvokeOn = 'click';
-  this.invokeOn        = this.elem.data('romo-ajax-invoke-on');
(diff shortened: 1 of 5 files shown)


assets/js/romo/ajax.js
```diff
@@ -1,26 +1,29 @@
...
-  this.elem       = $(element);
-  this.targetElem = $(this.elem.data('romo-ajax-target'));
+  this.elem = element;
+
+  var targetElemSelector = Romo.data(this.elem, 'romo-ajax-target');
+  if(targetElemSelector !== undefined) {
+    this.targetElem = Romo.f(targetElemSelector)[0];
+  } else [
+    this.targetElem = undefined;
+  }
```
kellyredding: @jcredding we discussed this, but I don't think you need to do this undefined check - the f/find meths should return an empty array if passed null/undefined on their own.

kellyredding (approved):
@jcredding one comment - otherwise I'm good here 👎

PR #149, opened 2017-10-13
Update `RomoCurrencyTextInput` component to not use jquery

jcredding:
This updates the `RomoCurrencyTextInput` component to not use
jquery. This is part of the effort to no longer require jquery to
use Romo. This removes all of the jquery usage within the
`RomoCurrencyTextInput` component except for how it initializes
an indicator text input for its elem. This will be updated when
the indicator text input is updated to no longer use jquery. None
of the other components initialize a currency text input
component.

Note: This didn't update any event names because they are already
prefixed with `romoCurrencyTextInput`.


kellyredding (approved):
@jcredding this looks good to me man 👎

PR #150, opened 2017-10-13
Update `RomoDatepicker` component to not use jquery

jcredding:
This updates the `RomoDatepicker` component to not use jquery.
`RomoDatepicker` component except for how it initializes an
indicator text input for its elem. This will be updated when the
indicator text input is updated to no longer use jquery. None of
the other components initialize a datepicker component.

`romoDatepicker` instead of just `datepicker`. This is part of
having everything properly namespaced in romo and ensuring it
should work without issue with other javascript libraries.


kellyredding (approved):
@jcredding changes look good 👎

PR #151, opened 2017-10-13
Update `RomoDropdown` component to not use jquery

jcredding:
This updates the `RomoDropdown` component to not use jquery. This
This removes all of the jquery usage within the `RomoDropdown`
component and also updates the datepicker, dropdown form, option
list dropdown, picker, select, and select dropdown components to
not use the jquery method for initializing a romo dropdown
component.

This removes using jquery's `unbind` to remove all event
handlers from elems. The `Romo.off` doesn't support unbinding all
event handlers but we also decided that Romo doesn't need to be
this aggressive with its components elems. This allows users to
bind event handlers to the same events that romo dropdown does
(though it's not recommended).

`romoDropdown` instead of just `dropdown` and renames variables as
well. This is part of having everything properly namespaced in
romo and ensuring it should work without issue with other
javascript libraries.

### DIFF
--- a/assets/js/romo/datepicker.js
+++ b/assets/js/romo/datepicker.js
@@ -95,39 +95,39 @@ RomoDatepicker.prototype.doBindDropdown = function() {
   if (this.elem.width() < 175) {
     this.elem.attr('data-romo-dropdown-width', '175px');
   }
-  this.romoDropdown = this.elem.romoDropdown()[0];
+  this.romoDropdown = new RomoDropdown(this.elem);
 
   this.romoDropdown.doSetPopupZIndex(this.elem);
   this.romoDropdown.bodyElem.addClass('romo-datepicker-calendar');
-  this.romoDropdown.elem.on('dropdown:popupOpen', $.proxy(this.onPopupOpen, this));
-  this.romoDropdown.elem.on('dropdown:popupClose', $.proxy(this.onPopupClose, this));
+  this.romoDropdown.elem.on('romoDropdown:popupOpen', $.proxy(this.onPopupOpen, this));
+  this.romoDropdown.elem.on('romoDropdown:popupClose', $.proxy(this.onPopupClose, this));
   this.romoDropdown.elem.on('blur', $.proxy(function(e) {
     this.blurTimeoutId = setTimeout($.proxy(function() {
       if (this.popupMouseDown !== true) {
-        this.romoDropdown.elem.trigger('dropdown:triggerPopupClose', []);
+        this.romoDropdown.elem.trigger('romoDropdown:triggerPopupClose', []);
       }
     }, this), 10);
   }, this));
   this.romoDropdown.elem.on('keydown', $.proxy(this.onElemKeyDown, this));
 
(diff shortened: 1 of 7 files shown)


assets/js/romo/datepicker.js
```diff
@@ -95,39 +95,39 @@ RomoDatepicker.prototype.doBindDropdown = function() {
   if (this.elem.width() < 175) {
     this.elem.attr('data-romo-dropdown-width', '175px');
   }
-  this.romoDropdown = this.elem.romoDropdown()[0];
+  this.romoDropdown = new RomoDropdown(this.elem[0]);
```
kellyredding: @jcredding just pass in `this.elem` as we discussed.

assets/js/romo/dropdown_form.js
```diff
@@ -1,7 +1,7 @@
 var RomoDropdownForm = function(element) {
   this.elem = $(element);
 
-  this.dropdown = this.elem.romoDropdown()[0];
+  this.romoDropdown = new RomoDropdown(this.elem[0]);
```
kellyredding: @jcredding `this.elem`

assets/js/romo/option_list_dropdown.js
```diff
@@ -176,7 +176,7 @@ RomoOptionListDropdown.prototype._bindElem = function() {
     this.optionFilterElem.trigger('indicatorTextInput:triggerIndicatorStop', []);
   }, this));
 
-  this.romoDropdown = this.elem.romoDropdown()[0];
+  this.romoDropdown = new RomoDropdown(this.elem[0]);
```
kellyredding: @jcredding `this.elem`

assets/js/romo/dropdown.js
```diff
@@ -88,90 +87,80 @@ RomoDropdown.prototype.doInitPopup = function() {
...
-  this.bodyElem.html('');
+  this.bodyElem.innerHTML = '';
   this.doInitBody();
   this.doPlacePopupElem();
-  this.elem.trigger('dropdown:loadBodyStart', [this]);
+  Romo.trigger(this.elem, 'romoDropdown:loadBodyStart', [this]);
 }
 
 RomoDropdown.prototype.doLoadBodySuccess = function(data) {
   Romo.initHtml(this.bodyElem, data);
```
kellyredding: @jcredding this needs updating.  This method no longer exists.  I think it should be `initUpdate` now.

assets/js/romo/dropdown.js
```diff
@@ -333,20 +334,20 @@ RomoDropdown.prototype.onResizeWindow = function(e) {
 }
 
 RomoDropdown.prototype.doPlacePopupElem = function() {
-  if (this.elem.parents('.romo-modal-popup').size() !== 0) {
-    this.popupElem.css({'position': 'fixed'});
+  if (Romo.parents(this.elem, '.romo-dropdown-popup').length !== 0) {
```
kellyredding: @jcredding this needs to stay `.romo-modal-popup`.  This is doing a check to see if it is a dropdown in a modal.

kellyredding (approved):
@jcredding a few comments - otherwise 👎

PR #152, opened 2017-10-13
Fix binding and unbinding to events on the window/document

jcredding:
This fixes binding and unbinding to events on the window/document.
This is part of an effort to no longer use jquery. This updates
the `_el` function to detect if the elem is the `window` or
`document` and if so it doesn't try to read or write an attribute
on them because they don't support the attribute methods. Instead
it uses a hard-coded identifier (`"window"` or `"document"`) for
them. This allows them to work with the event binding and ensures
we can bind and unbind events as needed.

@kellyredding - Ready for review. I rebased some stuff and was testing that the event binding worked on window and didn't need anything special done and saw this error.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -674,9 +674,13 @@ Romo.prototype._eid = 1;
 Romo.prototype._fid = 1;
 
 Romo.prototype._el = function(elem) {
-  elem._romoeid || (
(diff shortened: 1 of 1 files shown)


kellyredding (approved):
@jcredding great catch man!!  I assume you've tested this fix out in the console, yes?  If so, I'm good with this 👎

PR #153, opened 2017-10-13
Update `RomoModal` component to not use jquery

jcredding:
This updates the `RomoModal` component to not use jquery. This
This removes all of the jquery usage within the `RomoModal`
component and also updates the modal form to not use the jquery
method for initializing a romo modal component.

This removes using jquery's `unbind` to remove all event handlers
from elems. The `Romo.off` doesn't support unbinding all event
handlers but we also decided that Romo doesn't need to be this
aggressive with its components elems. This allows users to bind
event handlers to the same events that romo dropdown does (though
it's not recommended).

This also updates the event names to be prefixed with `romoModal`
instead of just `modal` and renames variables as well. This is

This also updates the `RomoDropdown` to not set `innerHTML` and
fixes a couple of its event bindings. These were missed in PR 151.

See #151

### DIFF
--- a/assets/js/romo/dropdown.js
+++ b/assets/js/romo/dropdown.js
@@ -48,12 +48,12 @@ RomoDropdown.prototype.doInitPopup = function() {
   var popupParentElem = Romo.closest(this.elem, Romo.data(this.elem, 'romo-dropdown-append-to-closest') || 'body');
   Romo.append(popupParentElem, this.popupElem);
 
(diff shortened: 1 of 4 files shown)


kellyredding (approved):
@jcredding this looks ok to me man 👎

jcredding:
@kellyredding - FYI, I updated the `innerHTML` calls for modal and dropdown in here (related to your comment in #157)

PR #154, opened 2017-10-13
Update `RomoTooltip` component to not use jquery

jcredding:
This updates the `RomoTooltip` component to not use jquery. This
This removes all of the jquery usage within the `RomoTooltip`
romo tooltip component.

`romoTooltip` instead of just `tooltip` and renames variables as
well. This is part of having everything properly namespaced in
romo and ensuring it should work without issue with other
javascript libraries.

### DIFF
--- a/assets/js/romo/tooltip.js
+++ b/assets/js/romo/tooltip.js
@@ -1,39 +1,39 @@
 var RomoTooltip = function(element) {
-  this.elem = $(element);
+  this.elem = element;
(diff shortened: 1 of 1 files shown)


kellyredding (approved):
@jcredding this looks fine as far as I can tell 👎

jcredding:
@kellyredding - FYI, tooltip didn't use `innerHTML` so nothing to update from your comment in #157.

PR #155, opened 2017-10-13
Update `RomoIndicatorTextInput` component to not use jquery

jcredding:
This updates the `RomoIndicatorTextInput` component to not use
jquery. This is part of the effort to no longer require jquery to
use Romo. This removes all of the jquery usage within the
`RomoIndicatorTextInput` component and also updates the currency
text input, datepicker, and option list dropdown components to
not use the jquery method for initializing a romo indicator text
input component.

`romoIndicatorTextInput` instead of just `indicatorTextInput` and
renames variables as well. This is part of having everything
properly namespaced in romo and ensuring it should work without
issue with other javascript libraries.

### DIFF
--- a/assets/js/romo/indicator_text_input.js
+++ b/assets/js/romo/indicator_text_input.js
@@ -1,5 +1,5 @@
 var RomoIndicatorTextInput = function(element) {
-  this.elem = $(element);
+  this.elem = element;
 
   this.defaultIndicatorClass     = undefined;
   this.defaultIndicatorPaddingPx = 5;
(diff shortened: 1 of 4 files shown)


assets/js/romo/indicator_text_input.js
```diff
@@ -116,13 +117,13 @@ RomoIndicatorTextInput.prototype._placeIndicatorElem = function() {
     var indicatorPosition  = this._getIndicatorPosition();
 
     // add a pixel to account for the default input border
-    this.indicatorElem.css(indicatorPosition, indicatorPaddingPx+1);
+    Romo.setStyle(this.indicatorEle, indicatorPosition, indicatorPaddingPx+1);
```
kellyredding: @jcredding type, you misspelled "Elem" here.

kellyredding (approved):
@jcredding one comment - otherwise this looks good man 👎

PR #156, opened 2017-10-14
rework romo form js to not use jquery

kellyredding:
This is part of switching off of jQuery and on to vanilla Romo js.
This involved a number of changes in addition to the expected
syntax changes.  The diff is a little rugged b/c I chose to
reorder the methods.

* first off, I'm trying to make the only public methods be things
  that should be public.  This means no event handler methods and
  no methods not intended to be called externally.  For the form,
  this means only the standard `doInit` method and the `daSubmit`
  method remain public.  All other methods were converted to
  private methods.
* I renamed the vars tracking the form's indicator elems to
  "spinner" since we plan on renaming that component.  However, I
  didn't rename the event names.  I expect whoever reworks that
  component will do that part.
* I chose to remove the e undefined checks around calling
  prevent default and/or stop propagation.  I think it is safe to
  count on these being events since we have moved to only calling
  event handlers with events.  I'd like the code to error if not
  so we can trace what's up.
* I broke out an `_onTriggerSubmit` handler to handle just the
  trigger submit event. This differs from the `_onSubmitClick`
  handler in that it has to reduce its disabled/confirm checks
  to determine those values (where the submit click can just check
  disabled/confirm on the single target elem.  For both of these
  handlers, I switched to checking disabled before checking
  confirm - you shouldn't get to confirm if the elem is disabled.
* I had to rewrite the functions to get the form values and build
  the form data.  jQuery previously took care of this for us.
* I updated the non browser get submit to go ahead and build
  form data (no files) and call `Romo.param` on the form data to
  build the get query string.  I'm assuming this works since
  form data is an object but this needs testing.  This matches the
  `Romo.param` call the `Romo.ajax` method does for ajax get
  requests so I wanted to be consistent.
* since `Romo.ajax` doesn't support a `dataType` opt where we can
  set the data type (typically 'json'), I had to update the submit
  success/error handlers to parse json as needed.

@jcredding sorry the diff is so gnarly.  ready for review.  Check me on this stuff - it is kinda complex and we'll need to test carefully with examples on the edge-case stuff here.

### DIFF
--- a/assets/js/romo/form.js
+++ b/assets/js/romo/form.js
@@ -1,86 +1,42 @@
-var RomoForm = function(element, givenSubmitElement, givenIndicatorElements) {
-  this.elem = $(element);
-  this.defaultSubmitElem = this.elem.find('button[type="submit"], input[type="submit"], [data-romo-form-submit]');
-  this.submitElem = $(givenSubmitElement || this.defaultSubmitElem);
-  this.defaultIndicatorElems = this.elem.find('[data-romo-indicator-auto="true"]');
-  this.indicatorElems = $(givenIndicatorElements || this.defaultIndicatorElems);
-  this.changeSubmitElems = this.elem.find('[data-romo-form-change-submit="true"]');
-  this.onkeySubmitElems = this.elem.find('[data-romo-form-onkey-submit="true"]');
+var RomoForm = function(formElem, givenSubmitElems, givenSpinnerElems) {
+  this.elem = formElem;
 
-  this.defaultListValuesDelim = ',';
+  var defaultSubmitElems = Romo.find(
+    this.elem,
+    'button[type="submit"], input[type="submit"], [data-romo-form-submit]'
+  );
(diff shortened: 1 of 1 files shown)


assets/js/romo/form.js
```diff
@@ -1,134 +1,141 @@
...
-RomoForm.prototype._doCompleteSubmit = function() {
-  this.elem.trigger('form:submitComplete', [this]);
-  if (this.submitQueued === true) {
-    this._doSubmit();
-  } else {
-    this.submitRunning = false;
+RomoForm.prototype._onFormKeyPress = function(e) {
+  if (Romo.data(this.elem, 'romo-form-disable-keypress') !== true) {
+    var targetElem = e.target;
+    if (ttargetElem.nodeName.toLowerCase() !== 'textarea' && e.keyCode === 13 /* Enter */) {
```
jcredding: @kellyredding - `ttargetElem` typo here

jcredding (approved):
@kellyredding - I think all of this makes sense to me and otherwise looks like it matches what we previously had. I'm good with this :boom:

PR #157, opened 2017-10-16
Update `RomoInline` component to not use jquery

jcredding:
This updates the `RomoInline` component to not use jquery. This
This removes all of the jquery usage within the `RomoInline`
component and also updates the inline form to not use the jquery
method for initializing a romo inline component.

This removes using jquery's `unbind` to remove all event handlers
from elems. The `Romo.off` doesn't support unbinding all event
handlers but we also decided that Romo doesn't need to be this
aggressive with its components elems. This allows users to bind
event handlers to the same events that romo inline does (though
it's not recommended).

This also updates the event names to be prefixed with `romoInline`
instead of just `inline` and renames variables as well. This is

### DIFF
--- a/assets/js/romo/inline.js
+++ b/assets/js/romo/inline.js
@@ -1,29 +1,29 @@
 var RomoInline = function(element) {
-  this.elem        = $(element);
-  this.toggleElem  = $(this.elem.data('romo-inline-toggle'));
+  this.elem        = element;
+  this.toggleElem  = Romo.f(Romo.data(this.elem, 'romo-inline-toggle'))[0];
   this.dismissElem = undefined;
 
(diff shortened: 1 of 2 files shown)


assets/js/romo/inline.js
```diff
@@ -1,80 +1,79 @@
...
 }
 
 RomoInline.prototype.doInit = function() {
   // override as needed
 }
 
 RomoInline.prototype.doLoadStart = function() {
-  this.elem.html('');
-  this.elem.trigger('inline:loadStart', [this]);
+  this.elem.innerHTML = '';
```
kellyredding: @jcredding we discussed switching this to `Romo.updateHtml(this.elem, '');` (both here and in dropdown, modal, and tooltip too).

kellyredding (approved):
@jcredding one minor comment - otherwise 👎

PR #158, opened 2017-10-16
Update `RomoOnkey` component to not use jquery

jcredding:
This updates the `RomoOnkey` component to not use jquery. This
This removes all of the jquery usage within the `RomoOnkey`.

This also updates the event names to be prefixed with `romoOnkey`
instead of just `onkey` and renames variables as well. This is


kellyredding (approved):
@jcredding this looks good 👎

PR #159, opened 2017-10-16
onkey: add a delay feature; cleanup form and opt list dropdown usage

kellyredding:
This adds an optional delay feature to onkey.  If a delay data
attr is specified, onkey will delay for that many millisecs before
triggering the onkey event.  If another trigger happens before the
delay period is over, the delay is reset.  So this means the delay
will always be N secs after the *last* onkey trigger event.

This also cleans up the form and option list dropdown js to use
the new feature instead of doing this manually.

Finally, this cleans up a few style things to be up to our latest
conventions.  This includes making methods the should be
considered "private" to have the leading underscore convention.

@jcredding ready for review.


jcredding (approved):

PR #161, opened 2017-10-16
Update `RomoSortable` component to not use jquery

jcredding:
This updates the `RomoSortable` component to not use jquery. This
This removes all of the jquery usage within the `RomoSortable`
romo sortable component.

`romoSortable` instead of just `sortable`. This is part of having
everything properly namespaced in romo and ensuring it should work
without issue with other javascript libraries.

### DIFF
--- a/assets/js/romo/sortable.js
+++ b/assets/js/romo/sortable.js
@@ -1,5 +1,5 @@
 var RomoSortable = function(element) {
-  this.elem = $(element);
+  this.elem = element;
 
   this.draggableSelector = '[data-romo-sortable-item="true"]';
   this.handleSelector = '[data-romo-sortable-handle="true"]';
@@ -12,7 +12,7 @@ var RomoSortable = function(element) {
   this.doInitPlaceholder();
   this._resetGrabClasses();
 
-  this.elem.trigger('sortable:ready', [this]);
+  Romo.trigger(this.elem, 'romoSortable:ready', [this]);
 }
 
 RomoSortable.prototype.doInit = function() {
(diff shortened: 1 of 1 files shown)


assets/js/romo/sortable.js
```diff
@@ -12,80 +12,82 @@ var RomoSortable = function(element) {
...
   // IE fix
   try {
     // FF fix, it won't drag without some data being set
-    e.originalEvent.dataTransfer.setData('text/plain', null);
+    e.dataTransfer.setData('text/plain', null);
   } catch(e) {}
 
-  this.draggedElem = $(e.target);
-  this.draggedElem.addClass(this.draggingClass);
+  this.draggedElem = Romo.f(e.target)[0];
```
kellyredding: @jcredding don't think you need to do this here, instead just `this.draggedElem = e.target;`

assets/js/romo/sortable.js
```diff
@@ -96,54 +98,59 @@ RomoSortable.prototype.onDragStart = function(e) {
...
+    elems = Romo.children(Romo.parent(this.draggedOverElem));
+    var draggedOverIndex = elems.indexOf(this.draggedOverElem);
+    if (placeholderIndex < draggedOverIndex) {
       this.dragDirection = 'down';
     } else {
       this.dragDirection = 'up';
     }
     var insertMethod = this.dragDirection === 'down' ? 'after' : 'before';
-    this.draggedOverElem[insertMethod](this.placeholderElem);
+    Romo[insertMethod](this.draggedOverElem, this.placeholderElem);
```
kellyredding, line 150: @jcredding what is `Romo[insertMethod]` here?

kellyredding (changes requested):
@jcredding a comment and a question - I'll look again.

assets/js/romo/sortable.js
jcredding: @kellyredding - You can access functions/attributes/values whatever you want to call them via `[]` on javascript objects. `Romo` is just a javascript object so `Romo['after']` or `Romo['before']` is the same as `Romo.after` and `Romo.before`.

kellyredding: @jcredding ahh yes, that's right.  Thanks for clarifying this man.

kellyredding (approved):
@jcredding OK, fix that one comment and I'm good here 👎

PR #163, opened 2017-10-16
romoDropdown: cleanups to get up to modern conventions

kellyredding:

* I updated the `setTimeout` calls to just user `1` for the
  timeout value as this was just intended to "kick" the reactor
  (not actually delay 100ms).
* I updated the parent check to use a `reduce` statement to find
  if `elemIsParentOfTarget`.  This seemed more appropriate than
  the prev manual implementation.

@jcredding ready for review.


jcredding (approved):
@kellyredding - Makes sense, good stuff :boom:

PR #165, opened 2017-10-16
romoModal: cleanups to get up to modern conventions

kellyredding:

* I updated the `setTimeout` calls to just user `1` for the
  timeout value as this was just intended to "kick" the reactor
  (not actually delay 100ms).
* I updated the `_dragStart` method to not use jQuery logic - this
  was missed in PR 151.
* I updated this to trigger a `romoModal:popupClosedByEsc` event
  like the dropdown does on Esc.  This is just to be consistent
  with romoDropdown's implementation.
* I added the `popupOpen()` and `popupClosed()` property methods
  and switched to using them in the implementation.  I mostly did
  this to be consistent with RomoDropdown's implementation but it
  is also a nice cleanup.

See #151 for reference.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/modal.js
+++ b/assets/js/romo/modal.js
@@ -1,41 +1,134 @@
-var RomoModal = function(element) {
-  this.elem = element;
-  this.doInitPopup();
-  this.romoAjax = new RomoAjax(this.elem);
-  this.romoAjax.doUnbindElem(); // disable auto invoke on click
-
-  if (Romo.data(this.elem, 'romo-modal-disable-click-invoke') !== true) {
-    Romo.on(this.elem, 'click', Romo.proxy(this.onToggleClick, this));
-  }
-  Romo.on(this.elem, 'romoModal:triggerToggle', Romo.proxy(this.onToggleClick, this));
-  Romo.on(this.elem, 'romoModal:triggerPopupOpen', Romo.proxy(this.onPopupOpen, this));
-  Romo.on(this.elem, 'romoModal:triggerPopupClose', Romo.proxy(this.onPopupClose, this));
-  Romo.on(this.elem, 'romoAjax:callStart', Romo.proxy(function(e, romoAjax) {
-    this.doLoadBodyStart();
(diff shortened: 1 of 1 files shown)


assets/js/romo/modal.js
```diff
@@ -110,153 +220,74 @@ RomoModal.prototype.doResetBody = function() {
...
-  e.stopPropagation();
-  this.doDragStart(e);
+  this._dragStart(e);
   return false;
 }
 
-RomoModal.prototype.doDragStart = function(e) {
+RomoModal.prototype._dragStart = function(e) {
   this.dragElem.addClass('romo-modal-grabbing');
   this.dragElem.removeClass('romo-modal-grab');
```
jcredding: @kellyredding - Be sure to update these as well; thanks for fixing what I missed.

assets/js/romo/modal.js
```diff
@@ -110,153 +220,74 @@ RomoModal.prototype.doResetBody = function() {
...
-  this._dragDiffX = e.clientX - this.popupElem[0].offsetLeft;
-  this._dragDiffY = e.clientY - this.popupElem[0].offsetTop;
-  $(window).on('mousemove', Romo.proxy(this.onMouseMove, this));
-  $(window).on('mouseup',   Romo.proxy(this.onMouseUp, this));
+  this._dragDiffX = e.clientX - this.popupElem.offsetLeft;
+  this._dragDiffY = e.clientY - this.popupElem.offsetTop;
+  Romo.on(window, 'mousemove', Romo.proxy(this._onMouseMove, this));
+  Romo.on(window, 'mouseup',   Romo.proxy(this._onMouseUp,   this));
 
   this.elem.trigger("romoModal:dragStart", [this]);
```
jcredding: @kellyredding - Update this as well please

jcredding (approved):
@kellyredding - A couple comments about logic needing to be updated in the drag start. Otherwise I'm good with this :boom:

PR #167, opened 2017-10-17
romoTooltip: cleanups to get up to modern conventions

kellyredding:
This also organizes the methods to match the dropdown and modal
components.

* I added the `popupOpen()` and `popupClosed()` property methods
  and switched to using them in the implementation.  I mostly did
  this to be consistent with RomoDropdown's implementation.

@jcredding ready for review.


jcredding (approved):

PR #168, opened 2017-10-17
romoInline: cleanups to get up to modern conventions

kellyredding:
I also reorganized the bind logic to make sure `doInit` was being

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/inline.js
+++ b/assets/js/romo/inline.js
@@ -1,28 +1,10 @@
-var RomoInline = function(element) {
-  this.elem        = element;
-  this.toggleElem  = Romo.f(Romo.data(this.elem, 'romo-inline-toggle'))[0];
-  this.dismissElem = undefined;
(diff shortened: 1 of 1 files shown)


assets/js/romo/inline.js
```diff
@@ -1,61 +1,79 @@
...
-RomoInline.prototype.doLoadStart = function() {
+RomoInline.prototype._loadStart = function() {
   Romo.updateHtml(this.elem, '');
   Romo.trigger(this.elem, 'romoInline:loadStart', [this]);
 }
 
-RomoInline.prototype.doLoadSuccess = function(data) {
+RomoInline.prototype._loadSuccess = function(data) {
   this.doShow();
   Romo.initHtml(this.elem, data);
```
jcredding: @kellyredding - I think I missed this one, should be `initUpdateHtml` I believe

jcredding (approved):

PR #169, opened 2017-10-17
Update `RomoSelectDropdown` component to not use jquery

jcredding:
This updates the `RomoSelectDropdown` component to not use jquery.
`RomoSelectDropdown` component and also updates the select
component to not use the jquery method for initializing a romo
select dropdown component.

`romoSelectDropdown` instead of just `selectDropdown` and renames
variables as well. This is part of having everything properly
namespaced in romo and ensuring it should work without issue with
other javascript libraries.


kellyredding (approved):
@jcredding nice job - this looks like I'd expect 👎

PR #170, opened 2017-10-17
romoSortable: cleanups to get up to modern conventions

kellyredding:
follow our leading-underscore convention.  It also does a few
style cleanups.

In addition, there were a few Romo js syntax fixes I noticed:

* In the `_bindDraggableElems` method, the events were being
  bound on collections of elems, which doesn't work in Romo js.
  I updated the event binding to go into the draggable elems
  for each loop so they would be bound on each elem individually.
* This removes an errant `index` method param from the
  `_resetGrabClasses` method.

@jcredding ready for review.


jcredding (approved):

PR #171, opened 2017-10-17
romoDatepicker: cleanups to get up to modern conventions

kellyredding:
follow our leading-underscore convention.  It also does a few
style cleanups and removes event checks before calling event
"stop" methods.

I noticed some goof ups in the new romo change html methods that
I cleaned up.  I also switched some manual `append` calls passing
in elems using the `elems` method, to just call `appendHtml` and
pass in the raw html instead.

Finally, I chose to removet he `doSetFormat()` method and the
`romoDatepicker:triggerSetFormat` event bind.  Instead, I made
`formatString()` a lazy property method which removes the need
for both of those calls as the format string is always looked
up live and never needs to be "set".

Note: this also update the dropdown/tooltip `doSetPopupZIndex`
methods to be public.  I got too aggressive making these methods
private in earlier cleanup efforts.  The datepicker expects to
call this method on its dropdown.  I made the tooltip's public
to keep it consistent with the dropdown.

@jcredding ready for review.  Check me and make sure those `append` changes make sense.  Thanks.

### DIFF
--- a/assets/js/romo/tooltip.js
+++ b/assets/js/romo/tooltip.js
@@ -118,6 +118,11 @@ RomoTooltip.prototype.doSetContent = function(value) {
   this.doPlacePopupElem();
 }
 
+RomoTooltip.prototype.doSetPopupZIndex = function(relativeElem) {
+  var relativeZIndex = Romo.parseZIndex(relativeElem);
+  Romo.setStyle(this.popupElem, 'z-index', relativeZIndex + 1100); // see z-index.css
+}
+
 // private
 
 RomoDropdown.prototype._bindElem = function() {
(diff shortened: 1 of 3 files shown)


assets/js/romo/tooltip.js
```diff
@@ -118,6 +118,28 @@ RomoTooltip.prototype.doSetContent = function(value) {
   this.doPlacePopupElem();
 }
 
+RomoTooltip.prototype.doSetPopupZIndex = function(relativeElem) {
+  var relativeZIndex = Romo.parseZIndex(relativeElem);
+  Romo.setStyle(this.popupElem, 'z-index', relativeZIndex + 1100); // see z-index.css
+}
+
+RomoTooltip.prototype._getPopupMaxAvailableHeight = function(position) {
```
jcredding: @kellyredding - Did you mean to move this method out of the private section?

jcredding (approved):
@kellyredding - One comment about the `_getPopupMaxAvailableHeight` method in tooltip, otherwise I'm good with this :boom:

assets/js/romo/tooltip.js
kellyredding: @jcredding no, I didn't great catch - also I didn't even mention these changes in the commit msg - oof!

PR #173, opened 2017-10-18
Update `RomoSelect` component to not use jquery

jcredding:
This updates the `RomoSelect` component to not use jquery. This
This removes all of the jquery usage within the `RomoSelect`
romo select component.

`romoSelect` instead of just `select`. This is part of having
everything properly namespaced in romo and ensuring it should work
without issue with other javascript libraries.

### DIFF
--- a/assets/js/romo/select.js
+++ b/assets/js/romo/select.js
@@ -1,5 +1,5 @@
-var RomoSelect = function(element) {
-  this.elem = $(element);
+var RomoSelect = function(elem) {
+  this.elem = elem;
 
   this.defaultCaretClass     = undefined;
   this.defaultCaretPaddingPx = 5;
@@ -10,21 +10,22 @@ var RomoSelect = function(element) {
 
   this.doSetValue(this._elemValues());
 
-  if (this.elem.attr('id') !== undefined) {
-    $('label[for="'+this.elem.attr('id')+'"]').on('click', $.proxy(function(e) {
+  if (Romo.attr(this.elem, 'id') !== undefined) {
+    var labelElem = Romo.f('label[for="'+Romo.attr(this.elem, 'id')+'"]')[0]
+    labelElem.on('click', Romo.proxy(function(e) {
       this.romoSelectDropdown.doFocus();
     }, this));
   }
 
-  $(window).on("pageshow", $.proxy(function(e) {
+  Romo.on(window, "pageshow", Romo.proxy(function(e) {
     this._refreshUI();
   }, this));
 
-  this.elem.on('select:triggerSetValue', $.proxy(function(e, value) {
+  Romo.on(this.elem, 'romoSelect:triggerSetValue', Romo.proxy(function(e, value) {
     this.doSetValue(value)
   }, this));
 
-  this.elem.trigger('select:ready', [this]);
+  Romo.trigger(this.elem, 'romoSelect:ready', [this]);
 }
 
 RomoSelect.prototype.doInit = function() {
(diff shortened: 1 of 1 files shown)


assets/js/romo/select.js
```diff
@@ -10,44 +10,40 @@ var RomoSelect = function(element) {
...
 RomoSelect.prototype.doSetValue = function(value) {
-  var values = undefined;
-  if (Array.isArray(value)) {
-    values = value;
-  } else {
-    values = [value];
-  }
-  values = values.filter($.proxy(function(v) {
-    return this.elem.find('OPTION[value="'+v+'"]')[0] !== undefined;
+  var values = Romo.array(value);
```
kellyredding: @jcredding note: this doesn't work for single values (which is what the old implementation was trying to handle).  We need to handle making `.array` work for single values in a sep effort to keep this implementation (which I'd like to do).

assets/js/romo/select.js
```diff
@@ -136,100 +132,100 @@ RomoSelect.prototype._bindSelectDropdown = function() {
...
-  var caretClass = this.elem.data('romo-select-caret') || this.defaultCaretClass;
+  this.caretElem = undefined;
+  var caretClass = Romo.data(this.elem, 'romo-select-caret') || this.defaultCaretClass;
   if (caretClass !== undefined && caretClass !== 'none') {
-    this.caretElem = $('<i class="romo-select-caret '+caretClass+'"></i>');
-    this.caretElem.css('line-height', parseInt(Romo.getComputedStyle(romoSelectDropdownElem[0], "line-height"), 10)+'px');
-    this.caretElem.on('click', $.proxy(this._onCaretClick, this));
-    romoSelectDropdownElem.append(this.caretElem);
+    this.caretElem = Romo.elems('<i class="romo-select-caret '+caretClass+'"></i>')[0];
+    Romo.css(this.caretElem, 'line-height', parseInt(Romo.css(romoSelectDropdownElem, "line-height"), 10)+'px');
```
kellyredding: @jcredding think you need `.setStyle` here.

assets/js/romo/select.js
```diff
@@ -136,100 +132,100 @@ RomoSelect.prototype._bindSelectDropdown = function() {
...
     // add a pixel to account for the default input border
-    this.caretElem.css(caretPosition, caretPaddingPx+1);
+    Romo.css(this.caretElem, caretPosition, caretPaddingPx+1);
 
     // left-side padding
     // + caret width
     // + right-side padding
     var dropdownPaddingPx = caretPaddingPx + caretWidthPx + caretPaddingPx;
-    romoSelectDropdownElem.css('padding-'+caretPosition, dropdownPaddingPx+'px');
+    Romo.css(romoSelectDropdownElem, 'padding-'+caretPosition, dropdownPaddingPx+'px');
```
kellyredding: @jcredding `.setStyle` here.

assets/js/romo/select.js
```diff
@@ -244,27 +240,31 @@ RomoSelect.prototype._setValues = function(newValues) {
...
-  setElems.prop('selected', true);
+  var unsetElems = unsetValues.map(Romo.proxy(function(value) {
+    return Romo.find(this.elem, 'OPTION[value="'+value+'"]')[0];
+  }, this));
+  var setElems = setValues.reduce(Romo.proxy(function(value) {
+    return Romo.find(this.elem, 'OPTION[value="'+value+'"]')[0];
+  }, this));
+
+  unsetElems.forEach(Romo.proxy(function(unsetElem) {
+    Romo.setAttr(unsetElem, 'selected', '');
```
kellyredding: @jcredding does this work for unsetting the DOM attr?  I thought this would result in `<option selected></option>`

kellyredding: @jcredding we saw that it doesn't, so we need to add a `.removeAttr` to the base js.  I'll do this in a sep effort.

kellyredding (changes requested):
@jcredding a few comments that need sep efforts.  I'll look again once this is rebased off of those.

jcredding:
@kellyredding - I've rebased this off of #176 and #177 and updated it with your feedback.

kellyredding (approved):
@jcredding changes look good man 👎

PR #176, opened 2017-10-18
add "rm" methods for removing attr/data/style

kellyredding:
This originally came up b/c we found we needed an "remove attr"
method to port the RomoSelect component off of jQuery.  Since we
needed that method anyway, I decided to add similar "rm" methods
for data and style as well.  I backported some existing code to
use the new `rmStyle` method as well.

Note: this also fixes an errant `setAttr` call on the option list
dropdown component.  This was missed when switching it off of
jQuery.

@jcredding ready for review.  This is needed to finish off #173.


jcredding (approved):

PR #177, opened 2017-10-18
Update `Romo.array` helper to cast non array-like values to arrays

jcredding:
This updates the `Romo.array` helper to cast non array-like values
to arrays. This is to avoid manually checking for arrays and
manually converting them to arrays if not.

The `Romo.array` helper now checks the value its passed and see
if it has a `length` property. If it has a length and can be
accessed like an array (via `[]`) then the object is array like
and is converted to an array. The helper also includes some
checks to avoid functions, strings, and the `window` object being
seen as an array like object because otherwise the helper would
try to convert them to an array. If the object is not array like
it is added as the single item in an array. Finally, if the value
is a `NodeList` or `HTMLCollection` then the helper includes a
short circuit to immediately convert them to an array. This
ensures the `Romo.f`, `Romo.find`, and `Romo.elems` helpers all
remain as fast as possible.

This also updates the select, picker, and the `Romo.triggerInitUI`
logic to no longer manually check for arrays and instead use the
`Romo.array` helper. This allows them to properly cast values to
arrays as needed.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -396,13 +396,7 @@ Romo.prototype.onInitUI = function(fn) {
 }
 
 Romo.prototype.triggerInitUI = function(onElems) {
-  var elems = undefined;
-  if (Array.isArray(onElems)) {
-    elems = onElems;
-  } else {
-    elems = [onElems];
-  }
-  elems.forEach(Romo.proxy(function(elem) {
+  Romo.array(onElems).forEach(Romo.proxy(function(elem) {
     this._initUICallbacks.forEach(function(fn) {
       setTimeout(function() {
         fn(elem);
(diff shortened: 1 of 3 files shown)


assets/js/romo/base.js
```diff
@@ -384,13 +384,7 @@ Romo.prototype.onInitUI = function(fn) {
...
 
 Romo.prototype.triggerInitUI = function(onElems) {
-  var elems = undefined;
-  if (Array.isArray(onElems)) {
-    elems = onElems;
-  } else {
-    elems = [onElems];
-  }
-  elems.forEach(Romo.proxy(function(elem) {
+  Romo.array(onElems).forEach(Romo.proxy(function(elem) {
```
kellyredding, line 399: @jcredding love this cleanup!

assets/js/romo/picker.js
```diff
@@ -36,12 +36,7 @@ RomoPicker.prototype.doInit = function() {
 }
 
 RomoPicker.prototype.doSetValue = function(values) {
-  var value = undefined;
-  if (Array.isArray(values)) {
-    value = values.join(this._elemValuesDelim());
-  } else {
-    value = values;
-  }
+  var value = Romo.array(value);
```
kellyredding: @jcredding need to keep the join here: add `.join(this._elemValuesDelim()`

kellyredding (approved):
@jcredding one comment, otherwise 👎 

I like this a lot (especially with that short-circuit for those methods that get called so often).

PR #180, opened 2017-10-18
base js: make event handling error if trying to use with elem collections

kellyredding:
This is a protection feature.  Since Romo's js api is so similar
to jQuery's, it is logical that a dev might accidentally think to
call `.on` passing a collection of elems and have it work.  This
won't work.  So this add's a loud error to let you know you are
not using Romo's api correctly.

This should also protect us from mistakes as we port Romo from
jQuery to vanilla js.

@jcredding ready for review.


jcredding (approved):

PR #181, opened 2017-10-18
Update `RomoInlineForm` component to not use jquery

jcredding:
This updates the `RomoInlineForm` component to not use jquery.
`RomoInlineForm` component. None of the other components
initialize or use the romo inline form component.

`romoInlineForm` instead of just `inlineForm`. This is part of
having everything properly namespaced in romo and ensuring it
should work without issue with other javascript libraries.

This also updates the `RomoForm` initialization, event names, and
variable names to match the updates to the `RomoForm` component.


kellyredding (approved):
@jcredding I'm good with this 👎

PR #182, opened 2017-10-18
update select/picker related components to get up to convention

kellyredding:
This is just a few small tweaks as these components are fairly
recent and were already mostly up to convention.  These components
actually established many of our modern conventions.  This mostly
just gets consistent "private" section comments and moves logic
into `_bindElem` where appropriate (as we've done in the other
components).

@jcredding ready for review.


jcredding (commented):

PR #183, opened 2017-10-18
base js: remove `getComputedStyle` method

kellyredding:
Now that we have the `.css` method (which is the exact same
implementation), we no longer need this method.  `css` is more
succinct and is what we'll be using going forward.

This removes the base js method and its usage in base js and
indicator text input.  The other uses have already been updated
as part of earlier cleanups/reworks in the effort to move off of
using jQuery.

@jcredding ready for review.


jcredding (approved):

PR #184, opened 2017-10-19
update modal/dropdown/inline form components to latest conventions

kellyredding:
This updates these form components to reorg their methods so that
the "private" methods all have the leading underscor convention
and follow the `_bindElem` pattern that our other compoenents use.

This also fixes an error from when these were converted to Romo
js syntax and off of jQuery syntax in previous efforts.  The form
elem isn't available on initial load b/c it is, most typically,
loaded via ajax "load success" events.  So the initial `_bindForm`
call won't find a form elem and the `.romoForm` property will be
undefined unitl the "load success" events happen.  This means we
need to alter the `_bindForm` logic to only do its stuff if there
is a form elem.  Previously jQuery would return a "collection"
whether a form elem existed or not so the event binding were safe
whether a form elem existed or not.  Not so anymore.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/dropdown_form.js
+++ b/assets/js/romo/dropdown_form.js
@@ -1,22 +1,9 @@
 var RomoDropdownForm = function(elem) {
   this.elem = elem;
 
-  this.romoDropdown = new RomoDropdown(this.elem);
-  this.doBindDropdown();
-
-  this.romoForm = undefined;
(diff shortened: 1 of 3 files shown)


jcredding:
@kellyredding - I'm good with this :boom: Nice catch on the form not always being there.

PR #185, opened 2017-10-19
form: convention updates and a couple of bugfixes

kellyredding:
This gets the method names more up to date with the rest of the
components, removing unnecessary `_do*` prefixes from the private
methods and switching to the `_bindElem` convention method name.

This also fixes a few syntax errors I found while testing the new
Romo js based logic out in the browser.

@jcredding ready for review.


jcredding:

PR #186, opened 2017-10-19
Cleanups from the switch to no longer use jquery

jcredding:
This is a small set of cleanups from the effort to stop using
jquery. This includes:

* Update the `parseInt` to always pass a radix of 10 to avoid
confusion and ensure it works consistently across browsers as
recommended by MDN.
* Update some remaining `$.proxy` calls. These were most likely
the result of a bad merge.
* Update a few final `attr` and `setAttr` calls that can use
`data` and `setData` instead.
* Update the datepicker `_highlightItem` to allow passing an
`undefined` value to it and update the `_onPopupClose` to pass
`undefined` instead of an empty jquery object (`$()`). This should
have been done when datepicker was updated to not use jquery but
was probably missed. This switches to passing `undefined` because
Romo doesn't support building empty objects like jquery did.
* Fix a couple missing parenthesis in the dropdown and modal
components.
* Fix modal form and inline form using the dropdown form for their
`_bindElem` function.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -244,7 +244,7 @@ Romo.prototype.parseZIndex = function(elem) {
 }
 
 Romo.prototype.parseElemZIndex = function(elem) {
-  var val = parseInt(this.css(elem, "z-index"));
+  var val = parseInt(this.css(elem, "z-index"), 10);
   if (!isNaN(val)) {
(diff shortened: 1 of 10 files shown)


kellyredding (approved):
@jcredding nice cleanups! 👎

PR #187, opened 2017-10-19
add `.updateText` base js method; switch to using it in components

kellyredding:
This is part of building out the Romo vanilla js api.  Since we
have an `.updateHtml` method for setting an elem's html content,
it feels more natural and consistent to have an `.updateText`
method for setting an elem's text content.

Note: I've chose to jux this next to the `.updateHtml` method def
since they are similar in function.  However, there isn't any
corresponding "init" method b/c text content doesn't need any
initialization.  Also, there isn't any other "text" manipulation
methods (like replace, before/after, prepend/append) b/c text
content can only be updated.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -298,6 +298,10 @@ Romo.prototype.updateHtml = function(elem, htmlString) {
   return this.update(elem, this.elems(htmlString));
 }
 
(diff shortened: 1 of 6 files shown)


jcredding (approved):

PR #188, opened 2017-10-19
Don't use selectors that start with `>`

jcredding:
This updates dropdowns, modals, and tooltips to not use a
selector that starts with `>`. This isn't a valid selector and
causes `querySelectorAll` to thow an error. To get around this,
this switches to using `Romo.children` and `Romo.is` to find any
children that match the class.

Note: There is a `:scope` pseudo class that can be used to make
a selector like `:scope > .className` but it doesn't have great
support in all the browsers.


kellyredding:
@jcredding works for me man 👎

PR #189, opened 2017-10-19
Update dropdown and modal to not always expect a `contentElem`

jcredding:
This updates the dropdown and modal components to not always
expect that they have content. This is particularly true when
they are initializing and they reset their body. Their
`contentElem` won't exist yet so styles can't be removed from it.
This also does a similar check for a `closeElem` and the modal's
`dragElem` which may not exist when the dropdown/modal is first
initialized.

Note: This doesn't update the tooltip component because it doesn't
follow the same pattern as dropdowns and modals. It doesn't have
a content elem and instead uses its `bodyElem` when resetting and
it seems to always have a `bodyElem`.


kellyredding (approved):
@jcredding nice cleanup! 👎

PR #190, opened 2017-10-19
Update append/prepend, before/after, and update to allow single elem

jcredding:
This updates the Romo helpers for updating or inserting elems to
allow passing a single elem. This is done by having them call
`Romo.array` on whatever they are passed and then using their
previous implementation which expected a set of elems. This was
noticed because dropdowns try to append their popup elem to a
popup parent and were erroring because it wasn't an array.

This also updates the `Romo.array` helper to short circuit if it's
passed an array object. This keeps the append/prepend,
before/after, and update fast when passed arrays of elems. This
probably should have been done in PR 177 when we originally
updated `Romo.array` to convert any value to an array and when
we added the short-circuit for `NodeList` and `HTMLCollection`.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -289,7 +289,7 @@ Romo.prototype.initReplaceHtml = function(elem, htmlString) {
 
 Romo.prototype.update = function(elem, childElems) {
   elem.innerHTML = '';
-  return childElems.map(function(childElem) {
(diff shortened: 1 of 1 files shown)


kellyredding (approved):
@jcredding thanks man - I like this.  We discussed adding in this a short-circuit in the `array` meth for `isArray`.  Add that and explain in the commit msg and I'm good here 👎

PR #191, opened 2017-10-20
Fix select displaying an empty text option as selected

jcredding:
This fixes the select component trying to display an empty text
option as the selected value. An empty text option is one whose
`innerText` is an empty string. This also fixes multi selects
display since they never display a selected value in the select
itself.

The component was previously displaying the `&nbsp;` as text and
not properly interpretting it as a space. To fix this, the
component now switches between using `updateHtml` and `updateText`
depending on if the selected option's `innerText` is empty.

This also fixes an issue with finding the selected option. The
logic forgot to selected the first selected option and instead
tried to read `innerText` from a collection.


kellyredding:
@jcredding good stuff 👎

PR #192, opened 2017-10-20
Fix dropdowns placing their popup elem

jcredding:
This fixes a couple issues with dropdowns placing their popup
elem. This is part of updating Romo to no longer use jquery.
These were found while trying to open a select component which
positions a dropdown on the bottom or top of the select input.

First, this fixes using `Romo.assign` with the `DOMRect` object
that the `getBoundingClientRect` returns. The `DOMRect` doesn't
work when passed to `Object.assign` (this is what `Romo.assign`
uses). It simply doesn't copy any of it's keys into the resulting
object. This caused the calculations to sometimes result in a
`NaN` value because they would add or subtract an `undefined`
value. This fixes the issue by avoiding using `Romo.assign` in
this case. There's not really any value in maintaining an offset
object because we can't pass it as a single arg like we previously
did with the jquery `offset` method. This switches to using
separate local variables and also updates variable names to try
and better manage all the different values.

Next, this updates the final `setStyle` calls to add the missing
`px` to the end of the values they are setting. Without the `px`
the values were being ignored which caused the dropdown to not
be positioned correctly.

### DIFF
--- a/assets/js/romo/dropdown.js
+++ b/assets/js/romo/dropdown.js
@@ -86,10 +86,16 @@ RomoDropdown.prototype.doPlacePopupElem = function() {
     Romo.setStyle(this.popupElem, 'position', 'fixed');
   }
 
-  var pos    = Romo.assign({}, this.elem.getBoundingClientRect(), Romo.offset(this.elem));
-  var w      = this.popupElem.offsetWidth;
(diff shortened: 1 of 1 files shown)


kellyredding:
@jcredding good stuff - I like the more explicit implementation you have here better anyway.  👎

PR #193, opened 2017-10-20
Update `ParentChildElems.remove` to ignore non element nodes

jcredding:
This updates the `ParentChildElems.remove` to ignore nodes that
are not element nodes. This fixes an issue with the `remove`
method trying to remove a text node. The `removeNodes` from the
mutation record can return non element nodes and then the parent
child elems logic tries to remove them. Since the `remove` method
requires element nodes this updates it to ignore any other kind
of nodes.


kellyredding:
@jcredding yep 👎  jquery was hiding this problem from us previously

PR #194, opened 2017-10-23
make modal/dropdown bind needed by romo-av public

kellyredding:
When previously changing the method organization and conventions
here, I was too aggressive and didn't realize romo-av relied on
calling these methods.  This switches to making the "public" and
removing the leading underscore convention.

I noticed this while working on rewriting romo-av in vanilla js.

@jcredding ready for review.


jcredding (approved):

PR #195, opened 2017-10-23
Bug fixes and updates to latest conventions for sortable component

jcredding:
This updates the sortable component to fix some bugs introduced
while updating it to no longer use jquery. This also includes
updates for the sortable component to make it match some of our
latest conventions. This is part of updating Romo to no longer
require jquery.

This fixes the `Romo.before` call when the dragged elem is
dropped. It was accidentally updated to insert the placeholder
elem before the dragged elem instead of inserting the dragged
elem before the placeholder elem. This was incorrectly updated
in PR 161 when jquery was removed. It previously used
`insertBefore` on the dragged elem and is passed the placeholder
elem. This is the opposite order for how the elems needed to be
passed to `Romo.before` which is what caused the issue.

This updates the dragging, drag over, and placeholder class to
be lazy methods and also removes defaulting their values. This
is because the `addClass` and `removeClass` helpers error when
using an empty string which was the previous default. Now the
sortable component checks if it has the class before trying to
add or remove it which is consistent with these being optional.
This makes the methods lazy to match our latest conventions to
avoid all data attrs having to be set when the component is
initialized.

This updates the sortable to use `undefined` instead of `null`
for its attributes that are only set when dragging an element. The
other romo components have centered around using `undefined` so
the sortable component should as well. This also updates a few of
the `if` to check for `undefined` instead of using falsey values.
This is more explicit and should help avoid logic seemingly
working with invalid values.

This also fixes a few typos on method names. These were
accidentally introduced when the methods were renamed to our
private nameing scheme.

@kellyredding - Ready for review. I believe this gets the sortable component fully working.

### DIFF
--- a/assets/js/romo/sortable.js
+++ b/assets/js/romo/sortable.js
@@ -4,8 +4,12 @@ var RomoSortable = function(elem) {
   this.draggableSelector = '[data-romo-sortable-item="true"]';
   this.handleSelector    = '[data-romo-sortable-handle="true"]';
 
-  this.draggedElem     = this.draggedIndex  = this.draggableSelected = null;
-  this.draggedOverElem = this.dragDirection = this.lastY             = null;
+  this.draggedElem       = undefined;
+  this.draggedIndex      = undefined;
+  this.draggableSelected = undefined;
+  this.draggedOverElem   = undefined;
+  this.dragDirection     = undefined;
+  this.lastY             = undefined;
 
   this.doInit();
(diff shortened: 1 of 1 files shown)


kellyredding:
@jcredding "shoudl" in commit msg.

kellyredding (approved):
@jcredding looks good - I like the more explicit nature of this and what we are moving to in general 👎

PR #197, opened 2017-10-23
Update the add/change HTML methods to error when passed non HTML

jcredding:
This updates the add/change HTML methods on `Romo` to error if
they are passed non HTML strings. This is to help with the switch
from jquery to vanilla js. In jquery, the `html` helper would take
either HTML or text and update the element(s) appropriately. With
Romo if the add/change HTML methods were passed strings that
didn't include HTML then they would either add nothing and if
it was the replace or update it would remove the elem or its
content. To avoid this, the methods now throw an error if the
parsed HTML is empty but the HTML string wasn't empty.

This also includes updating the add/change HTML methods to not
accept non string values like `undefined` and `null`. Previously
these would either remove an elem or its content when using the
replace or update helpers and they would do nothing when using
the other helpers (prepent/append/before/after). Since the
behavior is inconsistent between the helpers and could hide
problems we decided it would be better for these to error as well.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -283,7 +283,11 @@ Romo.prototype.replace = function(elem, replacementElem) {
 }
 
 Romo.prototype.replaceHtml = function(elem, htmlString) {
-  return this.replace(elem, this.elems(htmlString)[0]);
+  var replacementElem = Romo.elems(htmlString)[0];
+  if (replacementElem === undefined && (typeof(htmlString) !== 'string' || htmlString.trim() !== '')) {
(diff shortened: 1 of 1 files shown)


kellyredding (changes requested):
@jcredding the local var names are no good `return this.replace(elem, elem);` and we need to make sure `htmlString` is a string before calling `trim` on it.  We discussed adding a string type check to the logic on each.  Ideally it would throw that same error if html string was not a string.

I'll look again.

jcredding:
@kellyredding - I've updated this with better local var names and to handle non string values as well.

kellyredding (approved):
@jcredding changes look good to me man 👎

PR #198, opened 2017-10-24
Update option list dropdown to properly display options

jcredding:
This fixes an issue with the option list dropdowns not displaying
its options when their `displayHtml` didn't contain any HTML
elements. This made it appear as if selects and pickers didn't
have any options in their dropdown. This fixes the issue by
checking if there are any HTML elements in the options
`displayHtml` and if not then the `displayHtml` is added as text
to the option. This allows the options API to remain unchanged
and fixes the options displaying correctly.


kellyredding (approved):
@jcredding I'm OK with this.  We discussed how the `&nbsp;` fallback needs to be addressed but I think we are doing that in a separate effort.

I'm good with this if so 👎

jcredding:
Notes: I addressed the nbsp issue in #199

PR #199, opened 2017-10-24
Fix `&nbsp;` usage for select and picker family of components

jcredding:
This updates the select, picker, select dropdown, and option list
dropdown components `&nbsp;` usage. With the changes to the
`updateHtml` method in PR 197 trying to call `updateHtml` with
just `&nbsp;` errors. This is because there are no HTML elements
in the nbsp string. To avoid this error the nbsp strings are now
wrapped in a span.

The previous usage of `updateHtml` seemed to work with the nbsp
string but was actually not working correctly. It seemed to work
because the select and picker elements it was checked on kept
their height with or without the nbsp. So the nbsp appeared to be
working correctly even though it wasn't actually being inserted
into the DOM (the `updateHtml` call was removing the content of
the elem when the nbsp string was passed to it). The errors from
PR 197 revealed this issue but it would have also become apparent
for elements that relied on the nbsp to give them height as well.

This also updates the picker setting its text elem to match the
implementation from the select. The select was previously updated
to in PR 191. This keeps the components consistent since their
logic is very similar and avoids the picker displaying the
string `"&nbsp;"` in its elem incorrectly.

See #191 and #197 

### DIFF
--- a/assets/js/romo/option_list_dropdown.js
+++ b/assets/js/romo/option_list_dropdown.js
@@ -221,7 +221,7 @@ RomoOptionListDropdown.prototype._buildListOptionElem = function(item) {
   var itemElem    = Romo.elems('<li data-romo-option-list-dropdown-item="opt"></li>')[0];
   var value       = item.value       || '';
   var displayText = item.displayText || '';
-  var displayHtml = item.displayHtml || item.displayText || '&nbsp;'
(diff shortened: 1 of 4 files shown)


kellyredding:
@jcredding "implementatino" in last paragraph of commit msg.

kellyredding (approved):
@jcredding good stuff man 👎 I like how thorough our implementations are becoming.

PR #200, opened 2017-10-24
Don't set `undefined` data attrs

jcredding:
This updates the select and picker components to not write
`undefined` data attrs. This was noticed while investigating
issues with selects dropdowns. I noticed the element had a lot
of data attrs with `undefined` values. This was happening
because the select was writing some of its data attrs using
the values of other data attrs. If the other data attrs weren't
set their `data` method returned `undefined`. In everywhere
outside of the select and picker components we first check if
the data attr is undefined before trying to set another data
attr. This updates the select and picker components to follow
this pattern.

Note: This didn't seem to be causing any issues because the `data`
method is smart and properly deserializes the `undefined` value.
This change is still valueable because in the HTML it looks like
an issue and its inconsistent with the rest of Romo. There are
some edge cases that this could effect if the data attrs were
being used in selectors but based on the data attrs I don't think
this is likely.

@kellyredding - Ready for review. I was investigating this as a cause for the dropdown heights being incorrect but as the commit says I don't think this was causing any problems just noisy.

### DIFF
--- a/assets/js/romo/picker.js
+++ b/assets/js/romo/picker.js
@@ -173,17 +173,24 @@ RomoPicker.prototype._bindOptionListDropdown = function() {
 RomoPicker.prototype._buildOptionListDropdownElem = function() {
   var romoOptionListDropdownElem = Romo.elems('<div class="romo-picker romo-btn" tabindex="0"><span class="romo-picker-text"></span></div>')[0];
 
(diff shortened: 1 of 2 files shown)


kellyredding:
@jcredding "This *change* is still *valuable* ..."

kellyredding (approved):
@jcredding good stuff 👎

PR #201, opened 2017-10-24
Fix using window height/width, use document elem client height/width

jcredding:
This fixes dropdowns, modals, and tooltips detecting the correct
window height (and width for modals). This was noticed because
dropdowns were not being correctly sized to fit within the window
when they were taller than the viewport. This is a bug from
switching from `$(window).height` to `window.innerHeight` (same
for width). When making the change I believed these were
equivalent but jquery actually uses
`document.documentElement.clientHeight` for the windows height
(same for width). The document elem client height is aware of the
scroll bars and removes their height from its. This fixes the
dropdowns only growing to the page height and should also fix
modals and tooltips positioning and sizing (they haven't been
fully tested yet).


kellyredding (approved):
@jcredding ok then.  thx jquery for masking this 👎

PR #202, opened 2017-10-24
Fixes and cleanups related to selects

jcredding:
This is a set of fixes and cleanups related to selects. These
were discovered while testing selects and getting them working.
This includes:

* Update the onkey trigger to not run the filter logic unless
the key is a text key code. This was accidentally removed in
PR 159. This keeps the filter code from running while using
the arrow keys, enter, or tab with the select.
* Fix the arrow keys logic by checking if there is a next
or prev elem. This is needed both when trying to find a next
or prev elem and to bail out if a next or prev elem couldn't
be found.
* Check if the list class is defined when building a romo
option list dropdown list elem. The `_buildListElem` private
function is used to build the outer list elem for the option
list dropdown and the sub list elems for optgroups. For the
outer list elem there is no list class but for the optgroup list
elems there is.
* Update `RomoIndicatorTextInput` to properly set its indicator
position style. This was missing the "px" which caused the
style to be ignored.

This also updates the next and prev elem logic to work if a
current element isn't highlighted in the dropdown list. This
allows dropping a multi select and immediately being able to
use the arrow keys. When the multi select is opened nothing is
highlighted so there wasn't a curr elem. This updates the next
to find the first elem in the list and the prev to find the
last elem in the list.

See #159

### DIFF
--- a/assets/js/romo/indicator_text_input.js
+++ b/assets/js/romo/indicator_text_input.js
@@ -117,7 +117,7 @@ RomoIndicatorTextInput.prototype._placeIndicatorElem = function() {
     var indicatorPosition  = this._getIndicatorPosition();
 
     // add a pixel to account for the default input border
-    Romo.setStyle(this.indicatorElem, indicatorPosition, indicatorPaddingPx+1);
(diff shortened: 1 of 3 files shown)


kellyredding:
@jcredding nice cleanups here man 👎

PR #204, opened 2017-10-25
Fixes and cleanups for tooltips

jcredding:
This is a set of fixes and cleanups for tooltips. These were
discovered while testing tooltips and getting them working
correctly. This includes:

* Add missing `px` when setting the tooltip `top` and `left` styles
to place the tooltip. This was also done for dropdowns in PR 192
and modals in PR 203. This was missed in tooltips when switching
from jquery to vanilla js.
* Update the logic to place the tooltip to be consistent with the
dropdown logic (PR 192). This avoids using a css object and
switches to separate local variables for determing the top and
left for the modal. Since we are no longer using the jquery offset
method which took an object there isn't any gain in building a css
object.
* Simplify setting the tooltip arrow position data attr. It's
always set to the same value as the config position local var so
this switches to setting it once after the config position is
determined.
* Fix a bug setting tooltip body HTML. The tooltip will now try
to detect if the body contains HTML and if so it will set it as
HTML otherwise it will set it as text.
* Update dropdown local vars to consistently order height/top vars
before width/left vars. This was noticed while updating the
tooltip logic.

### DIFF
--- a/assets/js/romo/dropdown.js
+++ b/assets/js/romo/dropdown.js
@@ -94,8 +94,8 @@ RomoDropdown.prototype.doPlacePopupElem = function() {
   var elemTop    = elemOffset.top;
   var elemLeft   = elemOffset.left
 
-  var popupOffsetWidth  = this.popupElem.offsetWidth;
   var popupOffsetHeight = this.popupElem.offsetHeight;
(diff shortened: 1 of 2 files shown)


kellyredding (approved):
@jcredding great cleanups here man 👎

PR #205, opened 2017-10-25
Fixes and cleanups related to pickers

jcredding:
This is a set of fixes and cleanups related to pickers. These
were discovered while testing pickers and getting them working
correctly. This includes:

* JSON parse the results from the Romo ajax calls. Jquery ajax
used to do this automatically but now it needs to be done
manually. This was done when setting a value and when handling
the romo ajax success event.
* Switch the reading and writing the picker display text to use
the Romo `data` and `setData` instead of `attr` and `setAttr`.
Romo's data methods just write attrs under the covers so they
can be safely used. This also removes some comments explaining
why `attr` and `setAttr` were used.
* Fix a typo in romo ajax for its `_invoke` method. This was
noticed while typing in the picker because it makes many requests
back to back.
* Fix an old jquery event binding for the picker label elem. This
was accidentally missed while switching from jquery to vanilla js.

### DIFF
--- a/assets/js/romo/ajax.js
+++ b/assets/js/romo/ajax.js
@@ -112,7 +112,7 @@ RomoAjax.prototype._onCallError = function(xhr, errorType, error) {
 RomoAjax.prototype._completeInvoke = function() {
   this._trigger('romoAjax:invoke', [this]);
   if (this.invokeQueued === true) {
-    this._doInvoke();
+    this._invoke();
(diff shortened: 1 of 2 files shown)


kellyredding (approved):
@jcredding ugh - there were a bunch of bugs in picker from our conversion.  Thanks for cleaning this up 👎

PR #209, opened 2017-10-26
Update dropdowns, modals, and tooltips to use `setTimeout` to place popup

jcredding:
This updates dropdowns, modals, and tooltips to use `setTimeout`
to place their popups. This fixes issues with the components
calculating their height/width appropriately when their content
is updated with components like pickers and selects. Both of these
change their HTML which causes their height to change. This in
turn causes the outer components height to change as well so
previous calculations may no longer be correct. By moving the
components into a `setTimeout` this allows the other components
time to adjust their HTML before placing the popup elem.

Note: This wasn't done on the popup open or resize event handler
for the components because it caused the dropdowns, modals, and
tooltips to have minor display issues. When the popup was opened
sometimes there would be a quick flash of the popup positioned
incorrectly. When resizing the animation wasn't smooth and it
would jitter as it adjusted its position based on the new window
size. This limits using the `setTimeout` to when the content of
the popup might change thus causing its height or width to
change.

### DIFF
--- a/assets/js/romo/dropdown.js
+++ b/assets/js/romo/dropdown.js
@@ -324,14 +324,14 @@ RomoDropdown.prototype._resetBody = function() {
 RomoDropdown.prototype._loadBodyStart = function() {
   Romo.updateHtml(this.bodyElem, '');
   this._bindBody();
-  this.doPlacePopupElem();
+  setTimeout(Romo.proxy(this.doPlacePopupElem, this), 1);
   Romo.trigger(this.elem, 'romoDropdown:loadBodyStart', [this]);
 }
 
 RomoDropdown.prototype._loadBodySuccess = function(data) {
   Romo.initUpdateHtml(this.bodyElem, data);
   this._bindBody();
-  this.doPlacePopupElem();
+  setTimeout(Romo.proxy(this.doPlacePopupElem, this), 1);
   Romo.trigger(this.elem, 'romoDropdown:loadBodySuccess', [data, this]);
 }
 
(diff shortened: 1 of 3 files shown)


assets/js/romo/dropdown.js
```diff
@@ -41,7 +41,9 @@ RomoDropdown.prototype.doPopupOpen = function() {
   }
 
   Romo.addClass(this.popupElem, 'romo-dropdown-open');
-  this.doPlacePopupElem();
+  setTimeout(Romo.proxy(function() {
```
kellyredding: @jcredding we discussed not wrapping in the proxied anonymous function and just adding directly.

kellyredding (approved):
@jcredding one comment - otherwise 👎

jcredding:
@kellyredding - After playing around with this in the UI I chose to remove using `setTimeout` in the components `doPopupOpen` functions. As mentioned in the commit message, this sometimes caused the popups to flash in the incorrect spot and then immediately adjust. By not doing the `setTimeout` the popups don't seem to do this. I revered the `doPopupOpen` changes and updated the commit message. I think using `setTImeout` should probably be limited to when the content of the popups is changed since this is what causes the height/width to change.

kellyredding:
@jcredding "*would* jitter"

kellyredding (approved):
@jcredding changes make sense and the writeup is good.  We discussed trying this implementation without the `Romo.proxy` to see if it works, but otherwise I'm good here 👎

jcredding:
@kellyredding - I tried removing the `Romo.proxy` from the set timeouts and it doesn't work.

PR #210, opened 2017-10-27
Fix parsing z-index from parent elems

jcredding:
This fixes parsing z-index values from parent elems. This was
noticed while testing a datepicker on a modal. The datepicker
wasn't properly parsing its parent modals z-index so its dropdown
was appearing behind the modal. This fixes the parent z-index
parsing by having it properly short-circuit the loop. Previously
we were using return which in javascript only returns out of the
loop and not out of the function as a whole. Now the function
checks if it has a non zero z-index value and if so it stops
parsing parent elems z-index values and returns the non-zero
parent z-index.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -239,15 +239,15 @@ Romo.prototype.parseZIndex = function(elem) {
   }
 
   // for the case where z-index is inherited from a parent elem
+  var pval = 0;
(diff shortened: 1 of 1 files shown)


kellyredding (approved):
@jcredding word 👎

kellyredding:
@jcredding also thanks for cleaning up my mess here.

PR #211, opened 2017-10-27
Fix `Romo.elems` working for tr, th, td, and other table related tags

jcredding:
This updates the `Romo.elems` logic to handle table related tags.
Previously, when passing HTML for a tr, th, td, and other table
related tags the `elems` helper would return an empty array. This
is because the trick of setting the HTML on the body of a new
document wouldn't actually add the HTML. The browsers won't allow
adding "invalid" HTML fragments and simply ignore the request. The
table related tags are invalid because they are not within a
`<table>` elem. This was noticed with the datepicker component
because they try to build a table for their calendars and build
the individual cells and rows before adding them all to the
table.

This fixes the issue by borrowing logic from jquery for properly
wrapping these HTML fragments within their correct tags. This
allows the content to be added and then read back out as we were
previously doing. The `elems` helper now tries to add the HTML
straight to the body and read it off and if it can't find any
HTML it then tries wrapping the content. If the helper can't find
a tag in the HTML string or if the tag isn't a known tag that
needs to be wrapped an empty array is returned because we know
adding the HTML to the body and reading it off already couldn't
find any HTML elements. If a tag can be found and we have an
entry for it in the wrap configuration then the wrap HTML is
used with the HTML string and written to the body. The helper
then descends through the wrap HTML elements to get to the HTML
string elements and returns them.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -270,7 +270,24 @@ Romo.prototype.elems = function(htmlString) {
   context.head.appendChild(base);
 
   context.body.innerHTML = htmlString;
-  return this.array(context.body.children);
+  if (context.body.children.length !== 0) {
+    return this.array(context.body.children);
+  } else {
+    var results = Romo._elemsTagNameRegEx.exec(htmlString);
+    if (!results){ return []; }
+
+    var tagName = results[1].toLowerCase();
+    var wrap    = Romo._elemsWrapMap[tagName];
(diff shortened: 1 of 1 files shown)


kellyredding (approved):
@jcredding yeah, this makes sense.  Nice writeup.  We discussed moving the private helpers down to the private helpers section of the base js.

Do that and I'm good here 👎

PR #212, opened 2017-10-27
Fixes and cleanups related to inlines

jcredding:
This is a set of fixes and cleanups related to inlines. These were
discovered while testing inlines and getting them working
correctly. This includes:

* Update `doBindDismiss` calls to use the new function name,
`_bindDismiss`. This also removes an extra call in the
constructor. The bind dismiss is already being called in the
`bindElem` which is called by the constructor.
* Support multiple dismiss elems. This updates the logic to look
for multiple dismiss elems and bind to their click events. This
also checks all of the dismiss elems to see if any of them are
configured with a confirm data attr or are disabled when handling
their click.

### DIFF
--- a/assets/js/romo/inline.js
+++ b/assets/js/romo/inline.js
@@ -4,7 +4,6 @@ var RomoInline = function(elem) {
   this.doInit();
   this._bindElem();
 
-  this.doBindDismiss();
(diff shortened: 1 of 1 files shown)


kellyredding (approved):
@jcredding we discussed doing this like the form js submit elems does its confirm logic (ie check disabled first).

Do that and I'm good here 👎

PR #213, opened 2017-10-27
Fixes and cleanups related to datepickers

jcredding:
This is a set of fixes and cleanups related to datepickers. These
were discovered while testing datepickers and getting them working
correctly. This includes:

* Update the `formatString` function to return its data attribute
value. This caused dates to not be displayed correctly in the
input.
* Fix removing a class from the highlight elem and update the
logic that checks if the elements exist to make sure they are
not `undefined`. This is more consistent with how we check if
an elem exists or not throughout romo.
* Switch to using `updateHtml` instead of `appendHtml` for the
next and prev elems. These are empty elems so update and append
are functionally the same. Using `updateHtml` matches the usage
of `updateText` if a icon class isn't specified.
* Fix a couple locations using the `setData` helper but still
specifying `data-` in the attr name. This would cause the attr
to be prefixed with `data-data-` and not work as intended.
* Update private function usage to match their new names with
leading underscores.

### DIFF
--- a/assets/js/romo/datepicker.js
+++ b/assets/js/romo/datepicker.js
@@ -17,7 +17,7 @@ var RomoDatepicker = function(elem) {
 }
 
 RomoDatepicker.prototype.formatString = function() {
-  Romo.data(this.elem, 'romo-datepicker-format') || this.defaultFormat;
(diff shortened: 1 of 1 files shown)


kellyredding (approved):
@jcredding good stuff here - thanks for cleaning up my private method name mess here 👎

PR #216, opened 2017-10-30
selects: update to handle zero options

kellyredding:
This updates the elem values logic to only fall back to selecting
the first option as the selected if there is at least one option.
This was erroring on selects with no options in their markup.

This also updates the refresh ui logic to only grab the inner text
of the selected option if their is an option elem selected.  This
was missed when we originally converted this logic from the jQuery
version.

@jcredding ready for review.


jcredding (approved):

PR #218, opened 2017-10-30
Remove `setTimeout` when auto init components, init first

jcredding:
This updates the `triggerInitUI` logic to no longer init
components in a `setTimeout`. This is to fix an issue with
components not initializing before javascript from `Romo.ready`.
Javascript in a `Romo.ready` function was running expecting
components were initialized and ready but they weren't. This is
because by using `setTimeout` all of the component init was
pushed on the reactor loop after any of the already added
`Romo.ready` functions. This removes the `setTimeout` call so
the components are initialized and available.


jcredding:
@kellyredding - Didn't see your hotfix, closing this.

PR #219, opened 2017-10-30
Fix spinner resetting HTML when stopped

jcredding:
This fixes spinners resetting their HTML when they are stopped.
They were previously using `updateHtml` but many times the
content they had replaced was simply the text of a button or an
nbsp. When this would happen `updateHtml` would error saying it
wasn't valid HTML. This avoids trying to detect if the content it
replaced is HTML and instead switches to writing `innerHTML` which
allows passing text, nbsp, or HTML.


kellyredding (approved):
@jcredding looks good :-1:

PR #220, opened 2017-10-30
Simplify `setTimeout` calls, don't create extra anonymous functions

jcredding:
This updates `setTimeout` calls in the dropdown and modal
components to not create unnecessary anonymous functions. This
is to try and be as performant as possible and avoid extra
function calls. This doesn't change any functionality.

### DIFF
--- a/assets/js/romo/dropdown.js
+++ b/assets/js/romo/dropdown.js
@@ -21,13 +21,9 @@ RomoDropdown.prototype.doInit = function() {
 
 RomoDropdown.prototype.doToggle = function() {
   if (this.popupOpen()) {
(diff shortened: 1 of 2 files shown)


kellyredding:
@jcredding this is good by me - why only modals and dropdowns - were these the only places we were creating extra anonymous functions like this?

Anyway, :-1:

kellyredding (approved):
@jcredding - oops should have "approved" this :-1:

jcredding:
@kellyredding - Yeah, I looked through all of the `setTimeout` calls and the only ones that could be simplified like this were in the modals and dropdowns.

PR #221, opened 2017-10-30
Update `Romo.ajax` and `Form` to handle additional scenarios

jcredding:
This updates the `Romo.ajax` helper and the `Form` component to
handle additional scenarios. These were noticed while testing
the components.

For the `Romo.ajax` helper this updates it to always expect an
object as the data passed to it. This was previously the case
for `GET` requests because they passed the data to `Romo.param`
which only accepts key-value objects. This now updates non GET
requests to do similar logic by expecting to iterate over the
data and build a `FormData`. This fixes issues with non GET
calls that were passing key-value objects. The previous logic was
ignoring the key-value objects because the xhr `send` only accepts
specific classes. Note: The logic for building `FormData` was
pulled from the `Form` component.

This also fixes `Romo.ajax` calls that don't include a success
or error callback. This adds checks to make sure they are present
before trying to call them.

For the `Form` component this fixes issues with its non browser
`GET` submits trying to build params from a `FormData`. This
was erroring because `Romo.param` expects key-value objects and
errors when passed a `FormData`. This updates it to simply build
the form values and pass those onto the `Romo.param` helper.

This also updates the `Form` component to work with the
`Romo.ajax` expecting a key-value object by no longer passing it
`FormData`. The `Form` component no longer builds form data and
instead simply passes the form values return value. The form
values logic now also changes any keys whose values are single
item arrays to remove the array and just use the single item that
was in the array. This was causing servers to interpret the params
as arrays which caused server logic to not work as expected
because the params should have been non array values.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -561,13 +561,23 @@ Romo.prototype.decodeParamMap = [
 Romo.prototype.ajax = function(settings) {
   var httpMethod = (settings.type || 'GET').toUpperCase();
   var xhrUrl     = settings.url || window.location.toString();
-  var xhrData    = settings.data ? settings.data : null
-  if (xhrData && httpMethod === 'GET') {
-    var xhrQuery = Romo.param(xhrData);
-    if (xhrQuery !== '') {
-      xhrUrl = (xhrUrl + '&' + xhrQuery).replace(/[&?]{1,2}/, '?');
+  var xhrData    = settings.data ? settings.data : undefined
+  if (xhrData) {
+    if (httpMethod === 'GET') {
+      var xhrQuery = Romo.param(xhrData);
+      if (xhrQuery !== '') {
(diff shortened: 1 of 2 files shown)


assets/js/romo/base.js
```diff
@@ -560,13 +560,23 @@ Romo.prototype.decodeParamMap = [
...
+    if (httpMethod === 'GET') {
+      var xhrQuery = Romo.param(xhrData);
+      if (xhrQuery !== '') {
+        xhrUrl = (xhrUrl + '&' + xhrQuery).replace(/[&?]{1,2}/, '?');
+      }
+      xhrData = undefined;
+    } else {
+      var formData = new FormData;
+      for (var name in xhrData) {
+        Romo.array(xhrData[name]).forEach(function(value ){
```
kellyredding: @jcredding extra space after `value`

kellyredding (approved):
@jcredding one minor style comment but otherwise :-1:

We'll need to test this out carefully and make sure the edge cases work as before, but overall I think this is really cool.

PR #222, opened 2017-10-30
base: update `find` to optionally take a list of parent elems

kellyredding:
In this case it would find all elems under each parent elem in the
list.  This makes it more like jQuery's api and also eliminates
this need to manually iterate over each parent elem and redo that
same logic for each parent's elems.

This is part of updating certain base api methods to optionally
take and work with a list of given elems.  We are doing this to
not only make the api more friendly to work with, but also to make
common operations on sets of elems more performant.

@jcredding ready for review.


jcredding (approved):

PR #224, opened 2017-10-30
base: add init elems methods

kellyredding:
These methods are similar to the other `init*` methods except
they don't mutate the DOM.  You can pass it single or multiple
elems that are attached to the DOM or not.

Also, all the places we were calling `triggerInitUI` now call
`initElems`.  This is prep for making the trigger method a private
helper as part of reworking the init ui logic and API to simplify
it.  The `initElems` method is a better public api method as it
conforms to the naming convention of the other DOM methods.  No
matter how we init elems, we want this api in place.

@jcredding ready for review.


jcredding (approved):

PR #225, opened 2017-10-30
Update attr, style, and class helpers to allow multiple elems

jcredding:
This updates the attr, data, style, class, show/hide, and
scroll top/left helpers that modify an elem to allow being passed
multiple elems. This is part of an effort to make the helpers
more conveninent when dealing with collections of elems. This was
only done to the helpers that modify an elem because these
provide most of the convenience and avoid issues with the return
values of the other helpers. For example, if `attr` was passed a
collection of elems should it return the first elems attr or a
collection of all of the elems attrs? If it returns a collection
should passing a single elem also return a collection with one
item? If `hasClass` takes a collection should it be all of the
elems have the class or at least one of them has the class? To
avoid these issues the helpers that read or returned a single
value were left as-is.

This updates the helpers to use `Romo.array` on the elem or elems
that are passed to them. This will convert them to an array if
they aren't already by simply wrapping the single elem in an
array (see PR 223). The helpers then iterate over each elem and
apply the logic they would have normally only done to a single
elem.

See #223 

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -136,62 +136,79 @@ Romo.prototype.attr = function(elem, attrName) {
   }
 }
 
-Romo.prototype.setAttr = function(elem, attrName, attrValue) {
+Romo.prototype.setAttr = function(elems, attrName, attrValue) {
   var v = String(attrValue);
-  elem.setAttribute(attrName, v);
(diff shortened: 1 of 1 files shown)


kellyredding (approved):
@jcredding looks like I'd expect - thx 👎

PR #226, opened 2017-10-30
Update `Romo.remove` to take multiple elems

jcredding:
This updates the `Romo.remove` helper to take multiple elems. This
is part of an effort to make the helpers more conveninent when
dealing with collections of elems and avoid having extra noisy
`forEach` loops. This will allow passing a single elem or multiple
elems and remove them from the DOM.


kellyredding (approved):
@jcredding yep 👎

PR #227, opened 2017-10-30
base: rework elems init API

kellyredding:
We noticed, while evaluating updates to the api to task optional
elem collections, that the API to setup auto elem init components
was a bit cumbersome and involved a bit of boilerplate code.  This
reworks the API to make it so you only make a single api call
passing in the elem selector and the component class.  Romo now
handles all of the boilerplate init logic in it's private helpers.

This also renames the api, in general, as "elems init".  "UI" was
a little ambiguous and components may or may not be related to UI.
A better way to describe all of this is initializing elems with a
component.  The private helpers were also reworked to be slightly
more efficient now that the `find` method can take a set of elems.

Note: this also removes the `_addEventCallback` method.  This
method isn't used and should have been removed earlier but was
missed.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -1,5 +1,4 @@
 var Romo = function() {
-  this._initUICallbacks = [];
 }
 
 Romo.prototype.doInit = function() {
@@ -269,7 +268,7 @@ Romo.prototype.parseElemZIndex = function(elem) {
   return 0;
 }
 
-// DOM manipulation
+// elems init
 
 Romo.prototype.elems = function(htmlString) {
   var context = document.implementation.createHTMLDocument();
@@ -301,13 +300,19 @@ Romo.prototype.elems = function(htmlString) {
   }
 }
 
+Romo.prototype.addElemsInitSelector = function(selector, componentClass) {
+  this._elemsInitComponents[selector] = componentClass;
+}
+
 Romo.prototype.initElems = function(elems) {
-  return this.triggerInitUI(elems);
+  return this._elemsInitTrigger(Romo.array(elems));
 }
 
 Romo.prototype.initElemsHtml = function(htmlString) {
   return this.initElems(this.elems(htmlString));
 }
+
+// DOM manipulation
 
 Romo.prototype.remove = function(elem) {
   return elem.parentNode.removeChild(elem);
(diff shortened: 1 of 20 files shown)


assets/js/romo/base.js
```diff
@@ -881,6 +860,27 @@ Romo.prototype._elemsWrapMap = {
   'td':       [3, "<table><tbody><tr>", "</tr></tbody></table>"]
 };
 
+Romo.prototype._elemsInitComponents = {};
+
+Romo.prototype._elemsInitTrigger = function(onElems) {
+  for (var selector in this._elemsInitComponents) {
+    this._elemsInitFind(onElems, selector).forEach(Romo.proxy(function(initElem) {
+      new this._elemsInitComponents[selector](initElem);
```
jcredding: @kellyredding - I think you could pull out `this._elemsInitComponents[selector]` in a local var and use it in this function. This would allow dropping the `Romo.proxy` and avoiding the lookup in the loop. Not sure it's more efficient but maybe.

assets/js/romo/base.js
```diff
@@ -881,6 +860,27 @@ Romo.prototype._elemsWrapMap = {
...
+    }, this));
+  }
+}
+
+Romo.prototype._elemsInitFind = function(onElems, selector) {
+  var elems = Romo.find(onElems, selector);
+  onElems.forEach(function(onElem) {
```
jcredding: @kellyredding - You could do use `elems.concat` and `filter` to maybe shorten this? Something like:

```javascript
elems.concat(onElems.filter(function(onElem) {
  return Romo.is(onElem, selector);
});
```

I don't think it's any more efficient but its less lines.

jcredding (approved):
@kellyredding - Looks good :boom: I like the one liner in the components javascript. Made a few minor comments but I'm good with this either way.

assets/js/romo/base.js
kellyredding: @jcredding thanks for this suggestion.  I did what you suggested except I started with the onElems that matched and concatted the child elems that matched.  I wanted the parent elems that matched to be listed before the child elems so that the parents are initialized before their children?  That is why I previously did an `unshift`.  Make sense?

PR #228, opened 2017-10-31
Update `on`, `off, and `trigger` to take multiple elems

jcredding:
This updates the `on`, `off`, and `trigger` helpers to take
multiple elems instead of just a single elem. This is part of an
effort to make the helpers more convenient when dealing with
collections of elems and avoid having extra noisy `forEach` loops.

This updates the `on` helper to build a single proxy function and
bind it to the multiple elems. This avoids building a proxy
function for each elem but otherwise just loops over the previous
behavior in a `forEach` loop.

This also updates the `trigger` helper to build a single event and
then loop over the passed elems dispatch it. Similar to the `on`
helper this avoids building a new identical event for each elem
which is hopefully slightly more performant.

The also updates the `off` helper but it simply loops over its
elems and runs its previous logic.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -490,18 +490,9 @@ Romo.prototype.initAfterHtml = function(elem, htmlString) {
 
 // events
 
-Romo.prototype.on = function(elem, eventName, fn) {
-  var elemString = Object.prototype.toString.call(elem);
-  if (
-    elemString === '[object NodeList]'       ||
-    elemString === '[object HTMLCollection]' ||
-    Array.isArray(elem)
-  ) {
-    throw new Error('Can only bind events on individual elems, not collections.');
-  }
-
+Romo.prototype.on = function(elems, eventName, fn) {
   var proxyFn = function(e) {
-    var result = fn.apply(elem, e.detail === undefined ? [e] : [e].concat(e.detail));
+    var result = fn.apply(e.target, e.detail === undefined ? [e] : [e].concat(e.detail));
     if (result === false) {
       e.preventDefault();
       e.stopPropagation();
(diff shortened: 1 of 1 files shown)


kellyredding:
@jcredding "... helper to *build loop* over the ..."?  (did you just mean to say "loop"?)

assets/js/romo/base.js
```diff
@@ -478,49 +469,46 @@ Romo.prototype.on = function(elem, eventName, fn) {
...
-  });
-  this._handlers[key] = [];
+    var key = this._handlerKey(elem, eventName, proxyFn);
+    if (this._handlers[key] === undefined) {
+      this._handlers[key] = [];
+    }
+    // `addEventListener` doesn't register duplicates (same function on same
+    // elem) so we shouldn't store them in our handlers arrays either
+    if (this._handlers[key].indexOf(proxyFn) === -1) {
+      this._handlers[key].push(proxyFn);
```
kellyredding: @jcredding if duplicates aren't registered, why are we still pushing on to an array here?  Seems like we should just store one proxyFn per key (since the proxyFn is part of the key).

kellyredding (changes requested):
@jcredding I wanna discuss the whole duplicates thing with you.  I'll look again 👎

assets/js/romo/base.js
```diff
@@ -478,49 +469,46 @@ Romo.prototype.on = function(elem, eventName, fn) {
...
   proxyFn._romofid = this._fn(fn)._romofid;
 
-  elem.addEventListener(eventName, proxyFn);
-  var key = this._handlerKey(elem, eventName, proxyFn);
-  if (this._handlers[key] === undefined) {
-    this._handlers[key] = [];
-  }
-  this._handlers[key].push(proxyFn);
-}
+  elems.forEach(Romo.proxy(function(elem) {
```
kellyredding: @jcredding I think this needs to be `Romo.array(elems)`.  Missed this earlier

assets/js/romo/base.js
```diff
@@ -478,49 +469,46 @@ Romo.prototype.on = function(elem, eventName, fn) {
...
-  var elemString = Object.prototype.toString.call(elem);
-  if (
-    elemString === '[object NodeList]'       ||
-    elemString === '[object HTMLCollection]' ||
-    Array.isArray(elem)
-  ) {
-    throw new Error('Can only trigger events on individual elems, not collections.');
-  }
+Romo.prototype.off = function(elems, eventName, fn) {
+  elems.forEach(Romo.proxy(function(elem) {
```
kellyredding: @jcredding `Romo.array(elems)`.

assets/js/romo/base.js
```diff
@@ -478,49 +469,46 @@ Romo.prototype.on = function(elem, eventName, fn) {
...
+    (this._handlers[key] || []).forEach(function(proxyFn) {
+      elem.removeEventListener(eventName, proxyFn);
+    });
+    this._handlers[key] = [];
+
+  }, this));
+}
+
+Romo.prototype.trigger = function(elems, customEventName, args) {
+  elems.forEach(function(elem) {
```
kellyredding: @jcredding `Romo.array(elems)`.

assets/js/romo/base.js
```diff
@@ -458,16 +458,7 @@ Romo.prototype.initAfterHtml = function(elem, htmlString) {
...
-    elemString === '[object NodeList]'       ||
-    elemString === '[object HTMLCollection]' ||
-    Array.isArray(elem)
-  ) {
-    throw new Error('Can only bind events on individual elems, not collections.');
-  }
-
+Romo.prototype.on = function(elems, eventName, fn) {
   var proxyFn = function(e) {
     var result = fn.apply(elem, e.detail === undefined ? [e] : [e].concat(e.detail));
```
kellyredding: @jcredding did you test this at all?  I don't think this works b/c `elem` is not defined here.  Not sure what you wanna do about that.  I missed this earlier too.

kellyredding: @jcredding that is passing in `elem` as the value of `this` in the function (see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply).  Maybe we don't do this?  I don't think I even rely on `this` in event callbacks without using `Romo.proxy`.

assets/js/romo/base.js
jcredding: @kellyredding - It's exact duplicates so if you did:

```javascript
Romo.on([window, window], 'click', someFunc)
```

This would only register one event handler. If you call `Romo.on` twice though like:

```javascript
Romo.on(window, 'click', someFunc)
Romo.on(window, 'click', someFunc)
```

This would register two events listeners because they build different `proxyFn` in the `on` method. This is what happens when we have multiple dropdowns open on a page (via a select in a dropdown or something). Since they are separate `Romo.on` calls they register two different `proxyFn` functions that end up calling the same romo dropdown function.

jcredding:
@kellyredding - I've updated this like we discussed

kellyredding (approved):
@jcredding commit msg and code changes all look good - thanks man! 👎

PR #229, opened 2017-10-31
Update `Datepicker` to pass multi elems to `Romo.on`

jcredding:
This updates the `Datepicker` component to make use of romo's
helpers that now take multiple elems. This involved updating
the binding of mouse enter and click on the calendar items. This
was the only logic that could make use of the helper function
updates.

This also updates a `Romo.append` call passing the set of row
elems instead of iterating over the set and appending them one
at a time. This could have been done previously as the `append`
function has always taken multiple elems but was missed. I
noticed it because I was looking for loops used with romo helpers
to make use of the helpers now taking multiple elems.

This also adds a missing semicolon to the end of a line which is
part of our style conventions.


kellyredding (approved):
@jcredding word 👎

PR #230, opened 2017-10-31
Update `Form` to pass multi elems to `on` and `trigger`

jcredding:
This updates the `Form` component to make use of romo's helpers
that now take multiple elems. This involved updating the binding
of submit elems, change submit elems, and onkey submit elems and
also the triggering of events on its spinner elems. This was the
only logic that could make use of the helper function updates.


kellyredding:
@jcredding I think you mean "...  and *onkey* submit elems ..."

kellyredding (approved):
@jcredding one commit msg comment - otherwise 👎

PR #231, opened 2017-10-31
Update dropdown, modal, and inline to pass multi elems to helpers

jcredding:
This updates the dropdown, modal, and inline components to make
use of romo's helpers now accepting multiple elems. This involved
updating the binding of close/dismiss elems, binding modal drag
elems, and binding keyup on modals. This also updates the modals
to pass all of its drag elems when adding or removing classes
from them. This was all of the logic for these components that
could make use of the helper function updates.


kellyredding (approved):
@jcredding love these cleanups 👎

PR #232, opened 2017-10-31
Update `IndicatorTextInput` to pass multi elems to romo helpers

jcredding:
This updates the `IndicatorTextInput` component to make use of
romo's helpers that now take multiple elems. This involved adding,
removing, showing, and hiding the elem and indicator elem. This
was the only logic that could make use of the helper function
updates.

This also removes some private functions for showing or hiding
elems since they were just simple wrappers to calling romo's
`show` and `hide`.


kellyredding (approved):
@jcredding nice cleanups here man 👎

PR #233, opened 2017-10-31
Update select/picker related components to pass multi elems to romo helpers

jcredding:
This updates the select, picker, select dropdown, option list
dropdown, and selected options list components to make use of
romo's helpers that now take multiple elems.

This involved a number of small changes to each component:

* This updates the option list dropdown to pass all of its item
elems to `Romo.on` when binding to the mouse enter and click
events.
* This updates the select dropdown to pass the matching and non
matching elems from the word boundary filter when adding/removing
classes.
* This updates the selected options list to pass all of the elems
to remove to `Romo.remove` instead of removing them individually.
* This updates the select and picker components to not split their
class list and simply pass it to `addClass`. The `addClass` helper
was previously updated to accept a string of multiple class names.


kellyredding (approved):
@jcredding word man 👎

PR #235, opened 2017-11-01
picker/select/select-dropdown: properly bind label click events

kellyredding:
The previous implementation was looking for a label and binding
a click event handler.  However, it wasn't checking if there was
actually a label before trying to bind the event.  This switches
to just passing the set of elems returned by the find call to the
on event bind (now that this can take a set of elems).  For most
cases there will only be a single elem in the set and for the case
that there is no elem, no event will be bound.

This also fixes a small bug in the base `array` implementation.
We added a short-circuit for single element checks.  However this
broke the case where you pass `undefined` to `.array` b/c you
can't call `nodeType` on `undefined`.  This adds an undefined
check to the short-circuit to avoid this case.  This came up when
debugging the select/picker label error.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -700,10 +700,12 @@ Romo.prototype.array = function(value) {
     return Array.prototype.slice.call(value)
   }
 
-  // short circuit for passing elems, this ensures these remain fast and avoids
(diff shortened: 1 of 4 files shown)


jcredding (approved):

PR #236, opened 2017-11-01
currency text input: add onkey to update hidden input on key

kellyredding:
This makes it so you can type in the input and hit enter and have
it use the correct value.  Before, the correct value wasn't set
until a 'change' event happened on the input (which doesn't fire
when hitting enter on an input).

This brings in romo's onkey component on the input with no delay.
This means that on every keystroke, the hidden value is updated.
I changed the refresh function to take a skip param that skips
updating the elem value.  This is b/c you don't want the value
changing (ie reformatting) as the user types b/c it is
disorienting.  The onkey refresh skips this while the other don't.

Overall, this makes the input more friendly when typing and
hitting enter to submit forms.

@jcredding ready for review.


jcredding (approved):

PR #238, opened 2017-11-03
js: reorg event handler methods into event functions sections

kellyredding:
This is prep for reworking the event handler functions using a
new event function and component API.  This api will allow having
event handler functions that are specific to instances of
components.  This makes them safe to use with `Romo.off` calls.
Only the instance's function will unbound, not all components.

This just reorganizes the event handler functions so they are all
at the bottom of their component files.  I don't want the noise
of this to affect the diff when I apply the new API in a coming
effort.

@jcredding ready for review.


jcredding (approved):

PR #239, opened 2017-11-03
add API for defining RomoComponents

kellyredding:
This is a new class that commonizes logic that should be available
to all romo component objects.  This includes the standard
`doInit` constructor extension function definition and a new
"event function" api.

This goal here is to reduce the boilerplate code in each component
class definition and to provide common logic needed by all
components.

The new event function api allows defining event functions that
will be unique per instance of the class (even though they are
defined in our more typical style on the component's prototype.
We are choosing to provide this UI for just functions that are
used as event handlers.  This allows them to work properly with
`Romo.off` making it so you can unbind a single instances event
handler (when there are multiple on the page.

This new API is especially useful for dropdowns as it can be
common to have dropdowns open other dropdowns or have multiple
on the page at once.  Being able to unbind a single instances
event handler is powerful for these components.  It is similarly
powerful, although less commonly needed, for modals as well.

This includes the base API changes plus reworks to the dropdown
component for evaluation.  If we like this API, I'll update the
other components in a follow-up effort.

@jcredding ready for review

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -1,5 +1,4 @@
-var Romo = function() {
-}
+var Romo = function() {}
 
 Romo.prototype.doInit = function() {
   this.parentChildElems = new RomoParentChildElems();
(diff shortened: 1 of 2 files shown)


jcredding (approved):
@kellyredding - Looks like what we discussed, nicely done :boom:

PR #241, opened 2017-11-06
fix dropdown/tooltip placement in scrollable UI

kellyredding:
The current implementation didn't fully account for placing
popups correctly and keeping them placed correctly in UI that is
scrollable.  Previously, dropdowns/tooltips had checks to see if
they were nested in modals (which are position fixed) to make them
fixed as well.  This works for immediate child popups of modals
but doesn't work for granchild+ popups.  It also doesn't work if
the modal content overflows and can scroll.  It also doesn't work
for dropdowns/tootips in other non-modal but fixed UI.

This corrects these issues by adding more robust scroll behavior
and awaremess to dropdowns/tooltips:

* The popup stack now places all popups on both window resize and
  window scroll.  A new `placeAllPopups` function has been added
  to the stack to commonize the logic.  This method is also  aware
  of fixed popups.  By default it doesn't place fixed popups (ie
  modals).  You can,  However, you can pass in an options to force
  it to also place fixed popups.  The window resize event forces
  all popups while the scroll only places non-fixed popup.  This
  keeps modals that have been custom dragged in their custom
  position while scrolling.
* Romo base now has a `scrollableParents` method that returns all
  parents that can possibly be scrolled.  The dropdowns and
  tooltips use this method to attach scroll events to their
  scrollable parent elems.
* Tooltips now just close their popups on scroll to any of their
  scrollable parents or window.  The window has to be included
  since tooltips aren't part of the popup stack.  Closing is
  preferrable to dealing with edge-cases allowing tooltips to
  stay open and scroll as it is an equally viable behavior choice.
* Dropdowns tell the popup stack to place all popups on scroll to
  any of their scrollable parents.  Like the window scroll event
  the stack handles, this keeps the popups placed correctly as
  non window UI is scrolled.  This covers the case where a modal
  has dropdowns but also overflows its content and scrolls.

Finally, this also removes all hard-coded modal handling in the
dropdown/tooltip open popup logic.  The changes above mean that
these are no longer needed or apply whether in a modal or not.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -71,6 +71,16 @@ Romo.prototype.parents = function(childElem, selector) {
   }
 }
 
+Romo.prototype.scrollableParents = function(childElem, selector) {
+  return Romo.parents(childElem, selector).filter(function(parentElem) {
+    return (
+      Romo._overflowScrollableRegex.test(Romo.css(parentElem, 'overflow'))   ||
+      Romo._overflowScrollableRegex.test(Romo.css(parentElem, 'overflow-y')) ||
+      Romo._overflowScrollableRegex.test(Romo.css(parentElem, 'overflow-x'))
+    );
+  });
+}
+
 Romo.prototype.closest = function(fromElem, selector) {
   if (fromElem.closest) {
     return fromElem.closest(selector);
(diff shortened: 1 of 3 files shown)


jcredding:
@kellyredding - A few commit things:

* "awaremess" in 2nd paragraph
* "You can, However, you can" in first bullet

jcredding (approved):
@kellyredding - Looks like what we discussed, I'm good with this :boom:

PR #242, opened 2017-11-07
Update dropdowns/modals/tooltips to set max-height then position

jcredding:
This updates dropdowns/modals/tooltips to first set their content
`max-height` and then read their height and calculate their new
position. This is to fix an issue with dropdowns not always
positioning themselves correctly particularly when their content
continously changed. This specifically happened with a picker that
because of the number of results being returned would switch from
showing its results on top to the bottom of the picker. This
caused the dropdowns to sometimes not correclty position the
results above the picker and it would instead appear ontop of the
picker (as if it was stopped while moving to the top of the
picker).

This fixes the issue by changing how dropdowns place their elem.
Now, they first handle detecting their max height, then they
re-read their new height based on the max height that was just
set on the popup content and use it to calculate their position.
This ensures the last call to `doPlacePopupElem` properly uses
the latest height of the popup elem.

This also switches to using the `offsetHeight` instead of
parsing the css height as an integer when determining the max
height for a dropdown's/tooltip's content. The offset height is
the height of the elem including padding and borders which is
the same as the css height because we use border-box box sizing.
This switches to offset height because it's simpler and doesn't
require using `parseInt`.

This also changes the modal and tooltip to do the same change
even though there wasn't an issue seen with them. This keeps all
of the components as consistent as possible so they can all make
use of any fixes found with one of the others.

### DIFF
--- a/assets/js/romo/dropdown.js
+++ b/assets/js/romo/dropdown.js
@@ -38,23 +38,12 @@ RomoDropdown.prototype.doPopupClose = function() {
 }
 
 RomoDropdown.prototype.doPlacePopupElem = function() {
-  var elemRect   = this.elem.getBoundingClientRect();
-  var elemOffset = Romo.offset(this.elem);
-
-  var elemHeight = elemRect.height;
-  var elemWidth  = elemRect.width;
-  var elemTop    = elemOffset.top;
-  var elemLeft   = elemOffset.left
-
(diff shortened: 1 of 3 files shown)


kellyredding (approved):
@jcredding these look really nice - great cleanup! 👎

PR #243, opened 2017-11-07
option list dropdown: don't open on focus on clicks

kellyredding:
The open on focus logic caused dropdown to not open when clicked
while being closed.  The focus event triggers first and opens the
dropdown.  Then the click event triggers and toggles the dropdown
which closes it.

This fixes the bug by adding a `mousedown` event.  This event
triggers before `focus`.  On mousedown, we now always set the
open on focus state var to false as this is a click and the
click will handle opening the dropdown.

@jcredding ready for review.


jcredding:

PR #245, opened 2017-11-08
Update romo ajax, don't send empty form data

jcredding:
This updates the `Romo.ajax` function to check if the xhr data
it was passed is empty and if so the function won't try to use
it to build a query string or build a `FormData`. This
specifically fixes an issue with a `POST` sending empty `FormData`
and an HTTP server erroring because the request was invalid but
this also avoids attempting unnecessary logic to build a query
string that isn't needed.

Note: The error with sending empty `FormData` was only seen with
`POST` requests but didn't occur with `PUT` or `DELETE` requests.


kellyredding (approved):
@jcredding I'm good with this if it works in the various scenarios.  👎

PR #246, opened 2017-11-09
add `Romo.pushFn` method to formalize set timeout reactor push

kellyredding:
This is to add a formal name to the technique of using a timeout
to "push" a function to run at the end of the current reactor
stack of functions to run.  This has no behavior change - it only
formalizes a concept to express intent.  Also, if js ever adds a
formal way to do this without the set timeout hack, we only have
to update a single place.

@jcredding ready for review.


jcredding (approved):

PR #247, opened 2017-11-09
make add child elems to parents always run in a pushFn call

kellyredding:
This pattern was done in every usage of this API.  Instead of
repeating this (and the comments describing the purpose of it),
this commonizes the pattern in the RomoParentChildElems api.

There are no behavior changes here - just updates to remove some
code duplication.

@jcredding ready for review.


jcredding (approved):
@kellyredding - Nice cleanup :boom:

PR #248, opened 2017-11-16
pickers/selects: update to properly calc caret width/padding

kellyredding:
There were some bugs from the port to Romo js in previous efforts.
These went unnoticed b/c they positioned fairly close to what it
should be with the errors.

This switches setting the caret padding to include the 'px' in the
`Romo.setStyle` call.  JQuery was previously taking care of this
for us and we missed this in the port.

This also updates both components to have a default caret width.
In very edge-case scenarios the width of the caret elem can't be
determined.  This came up in our app b/c the parent of the select
was `display:none`.  This breaks Romo's and jQuery's width calcs.
This adds a default to fall back to in the case you can't parse
the width.  I chose 18px b/c this is close to FontAwesome's
`fa-fw` width with Romo's default font size.  In the case this
isn't a suitable default, you can always fallback to setting the
`data-romo-{picker|select}-caret-width-px` data attr.

@jcredding ready for review.

### DIFF
--- a/assets/js/romo/picker.js
+++ b/assets/js/romo/picker.js
@@ -3,6 +3,7 @@ var RomoPicker = RomoComponent(function(elem) {
 
   this.defaultCaretClass     = undefined;
   this.defaultCaretPaddingPx = 5;
+  this.defaultCaretWidthPx   = 18;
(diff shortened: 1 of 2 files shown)


jcredding (approved):

PR #250, opened 2017-11-30
add `romo-bg-striped` style class

kellyredding:
This is to just further fill out the bg style class offerings
since this is a formal color var/mixin.  This matches our other
bg style classes.

@jcredding ready for review.


jcredding (approved):

PR #251, opened 2017-11-30
Add a `Romo.height` and `Romo.width`, fix height/width calc in IE

jcredding:
This fixes issues in Internet Explorer with trying to calculate
the height/width of an element. This is part of an effort to
update romo to support later versions of IE. IE's
`getComputedStyle` for height and width isn't consistent with the
other browsers when the element is using border-box box sizing.
This causes many things to not be sized correctly.

This fixes the issue by using `getBoundingClientRect` to get the
height or width of an element. This is abstracted behind the
`Romo.height` and `Romo.width` helpers which will allow us to
switch to a different method for getting the height or width if
the need arises.

This updates any usage of `Romo.css` with width or height to
instead use the new helpers. For some logic this also required
adding `px` to the numbers returned by the width and height
helpers particularly when setting an element's style.

This also updates some manual usage of `getBoundingClientRect`
in the dropdown and tooltip components to also switch to the
`Romo.width` and `Romo.height` helpers. This is to centralize
this logic and if we need to change how we calculate height/width
then we can change it in one place for all of romo.

This also simplifies the picker and select setting the line-height
of their caret element. We were previously parsing the line-height
to an integer and then re-adding the `px` to it to set the style.
In this case we can just directly read the style from one elem
and set the style from the other elem and not worry about parsing
it into an integer.

### DIFF
--- a/assets/js/romo/base.js
+++ b/assets/js/romo/base.js
@@ -242,6 +242,14 @@ Romo.prototype.hide = function(elems) {
   });
 }
 
+Romo.prototype.height = function(elem) {
+  return elem.getBoundingClientRect().height;
+}
+
+Romo.prototype.width = function(elem) {
+  return elem.getBoundingClientRect().width;
+}
+
 Romo.prototype.offset = function(elem) {
   var elemRect = elem.getBoundingClientRect();
   var bodyRect = document.body.getBoundingClientRect();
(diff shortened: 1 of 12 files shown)


kellyredding (approved):
@jcredding this all looks good to me - really nice cleanups here with this change 👎

PR #252, opened 2017-12-07
Update Modal and DropDown `openPopup`

nweathersbee:
This updates the `openPopup` method in both the Modal and
Dropdown js. In both cases, the `contenElem` was undefined,
thereby erroring when `outerHTML` was being called on it. This
is erroring live in an app. This is now updated to correctly find 
the `contentElem`.

Also, I changed the `outerHTML` call on the `contentElem` to
`innerHtml`. With the "outer" call, nothing was being returned
as the standard mark-up for this situation has a `romo-display-none`
class. Changing to `innerHTML` fixed this issue.

This method, in both the Modal and Dropdown js, were updated in a
previous effort to get the `RomoModal` component to stop using
jquery. For reference see: 1e9381e

@redding  Ready for review.

### DIFF
--- a/assets/js/romo/dropdown.js
+++ b/assets/js/romo/dropdown.js
@@ -111,8 +111,8 @@ RomoDropdown.prototype._openPopup = function() {
   Romo.on(Romo.scrollableParents(this.elem), 'scroll', this._onScrollableParentsScroll);
 
   if (Romo.data(this.elem, 'romo-dropdown-content-elem') !== undefined) {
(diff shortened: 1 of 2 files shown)


jcredding (approved):
@nweathersbee - Changes look good but we need to adjust the commit message :boom:

kellyredding (approved):
@nweathersbee thanks man.  This looks good to me once you adjust the commit msg as Collin mentions. 👎

nweathersbee:
@kellyredding Updated commit msg per @jcredding.

PR #253, opened 2017-12-07
Update `doStart` and `doStop` to methods in Spinner js

nweathersbee:
This updates the `doStart` and `doStop` methods in the Spinner js
to include undefined checks before any of their specific logic
is run. This allows for calling start on an already started spinner
and stop on an already stopped spinner. This would error in some
cases when the same spinner elems were triggered to start/stop
multiple times.

@redding Ready for review.

### DIFF
--- a/assets/js/romo/spinner.js
+++ b/assets/js/romo/spinner.js
@@ -11,6 +11,10 @@ var RomoSpinner = RomoComponent(function(elem) {
 });
 
 RomoSpinner.prototype.doStart = function(customBasisSize) {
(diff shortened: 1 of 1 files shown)


kellyredding:
@nweathersbee I would change that last sentence to instead say "... *the same* spinner elems were triggered to start/stop *multiple times* ..."

kellyredding (approved):
@nweathersbee changes look good 👎

jcredding (approved):
@nweathersbee - Looks good :boom:

PR #254, opened 2017-12-07
fix a bug when adding popup stack elems

kellyredding:
Previously, this added the item to the stack and then allowed
the body click to propagate before opening the item.  In certain
scenarios, the body click could cause the stack to purge all items
thinking the click was a close all operation.

This switches to delay adding the item to the stack until after
the body click has propagated.  This should have been done before
and is a better design.  Now the item is added and immediately
opened - all after the body click has propagated.

@jcredding ready for review.

@nweathersbee FYI.


jcredding (approved):

PR #255, opened 2017-12-07
Fix form submit click checking if its submit is disabled or a confirm

jcredding:
This fixes the form submit click checking if the submit elem is
disabled or if its supposed to trigger a confirm. This sometimes
worked depending on if there were elements within the submit
elem. If elements were withing the submit elem sometimes the
`e.target` would be one of the elems within the submit instead of
the submit. This fixes the issue by checking if the target is a
submit and if not finding the closet submit. This ensures the form
correctly sees if a submit is disabled or a confirm.


kellyredding (approved):
@jcredding looks good man 👎

PR #256, opened 2017-12-08
Update Spinner js to delay running some of its logic

nweathersbee:
This updates the Spinner js to delay running some of its logic
in the `doStart` and `doStop` methods. This allows any events
triggered before the spinner start/stop calls to propagate thru.

Since the spinner changes its elem contents, if any previously
triggered events relied on those contents, the event would not
execute properly. This was noticed in one of our apps as a
function was called on the contents of elem that the spinner had
already changed, causing the function to perform erroneously.

@redding Ready for review.


kellyredding (approved):
@nweathersbee I'm good with this but I'd like to look at the commit msg after you've tweaked it like we discussed. 👎

jcredding (approved):
@nweathersbee - I'm good with this :boom:

PR #257, opened 2017-12-20
Add `RomoDate.daysDiff` helper

jcredding:
This adds a helper for calculating the number of days between
two dates. The helper converts the dates to their millisecond
time values, subtracts them, and then converts the difference to
a number of days by dividing it by the number of milliseconds in
a day. The helper rounds the difference to get an integer of the
days.

### DIFF
--- a/assets/js/romo/date.js
+++ b/assets/js/romo/date.js
@@ -78,6 +78,12 @@ var RomoDate = {
     return RomoDate.daysInMonth(d) - d.getDate() + 1;
   },
 
+  daysDiff: function(firstDate, secondDate) {
+    var fd = RomoDate.date(firstDate);
+    var sd = RomoDate.date(secondDate);
+    return Math.round((fd.getTime() - sd.getTime()) / 864e5); // 1000 * 60 * 60 * 24
+  },
+
   isoWeekNum: function(weekDate) {
     var d = RomoDate.date(weekDate)
 


assets/js/romo/date.js
```diff
@@ -78,6 +78,15 @@ var RomoDate = {
     return RomoDate.daysInMonth(d) - d.getDate() + 1;
   },
 
+  daysDiff: function(firstDate, secondDate) {
+    var diff = (firstDate.getTime() - secondDate.getTime()) / 864e5; // 1000 * 60 * 60 * 24
```
kellyredding: @jcredding did not know you could represent integers that way - cool!

kellyredding: @jcredding I'd prefer casting the given "dates" to actual dates like the method above does.  This keeps the api consistent and I think allows you to pass strings directly.

kellyredding (approved):
@jcredding a comment about casting but I'm good once that is done 👎

jcredding:
@kellyredding - FYI I switched to rounding the difference to get an integer for the days instead of flooring it. Javascript floating point precision seems to get off randomly so you get a value of like `25.98333333` days when it should be `26`. This seems to fix all the issues in our app but I'll keep an eye on it going forward.

kellyredding:
@jcredding ok, that makes sense.

PR #258, opened 2017-12-21
Update form to allow event submitting

jcredding:
This updates the form component to allow configuring it to do an
event submit instead of a browser or ajax submit. The goal is to
provide a way to use romo form but not have it make a request to
the server (either as an standard HTTP request or as an ajax
request).

This adds a data attr for event submitting and checks it whenever
the form is submitted (either via a click or the trigger event).
The event submit simply builds the form values the same as an
ajax submit then triggers an event submit event with the form
values. This will allow listening for the event and then using the
form values as needed. This also proxies the event in the dropdown
form, inline form, and modal form so it can be used with those
components as well.

This also renames the non-browser submit methods to use ajax
instead of non-browser. Since the submit logic is no longer a
binary between browser or non-browser this makes for a better
name (because event submitting is also non-browser submitting).
This also matches the other private method that is already
named `_ajaxSubmit`.

Finally, this also moves the browser submit event before
submitting the form. This is because submitting the form reloads
the page so an event after it can't really be used. This allows
the event to be used before the page reloads.

### DIFF
--- a/assets/js/romo/dropdown_form.js
+++ b/assets/js/romo/dropdown_form.js
@@ -89,6 +89,9 @@ RomoDropdownForm.prototype._bindForm = function() {
     Romo.on(formElem, 'romoForm:browserSubmit', Romo.proxy(function(e, romoForm) {
       Romo.trigger(this.elem, 'romoDropdownForm:romoForm:browserSubmit', [romoForm, this]);
     }, this));
+    Romo.on(formElem, 'romoForm:eventSubmit', Romo.proxy(function(e, formValues, romoForm) {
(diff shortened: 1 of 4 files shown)


kellyredding (approved):
@jcredding I'm good with this 👎

jcredding:
@kellyredding - FYI I updated this to proxy the new event submit on the dropdown form, inline form, and modal form. I didn't notice I needed this till I used it in a dropdown form.

kellyredding:
@jcredding good stuff - you might mention that in the commit msg.  Thanks for letting me know.

PR #261, opened 2018-03-02
update word boundary filter to use "items" api instead of "elems"

kellyredding:
This is more appropriate b/c the filter works with more than just
elems - any set of objects can be filtered using this.  Therefore
we switch to the more generic "items" interface and variable names.

This would be a breaking change, but I've kept the legacy
`.{matching|notMatching}Elems` methods for backwards compatibility.
Once we've migrated everything that uses this, we'll remove these
in a breaking change.

@jcredding ready for review.


jcredding (approved):

PR #262, opened 2018-03-02
update picker/select components to emit "multiple change" events

kellyredding:
I noticed this while working on the coming "color select"
component.  I noticed that if you clicked on a multiple selected
option, the value was removed but no change events were emitted.
This causes any handling meant to run on "change" to not fire when
removing multi items.

This updates the select/picker to emit "multiple change" events
when they add/remove items from their selected options list.
This is similar to how they proxy the option list dropdown's change
events.

The events emitted are the standard "change" event but also emit
the "multiple change" custom events.  These events are only
emitted when using the multiple option.  They emit with the
full set of items and either the item being added or removed
depending on the scenario.

@jcredding ready for review.


jcredding (approved):

PR #263, opened 2018-03-02
add the color select component

kellyredding:
This is a select that renders colored html for choosing colors.
Colors are defined by setting a json data attr.  Otherwise, it
is pretty much just like the picker.  Despite being named "select",
it should be used with a hidden/text input elem.  It is only a
"select" in its behavior.  You don't render nested option markup
or anything.

This supports all of the standard select/picker features: multi,
filtering, custom values, empty values, etc.  It proxies an
option list dropdown and a selected options list just like the
picker does.

In addition, there are a number of select/picker related cleanups
including setting actual html for the custom option display html
and other minor variable name changes and style tweaks I noticed
while working this up.  I also added an undefined check that I
ran into that errored when testing the component.

colorselect1

@jcredding ready for review.


assets/js/romo/color_select.js
```diff
@@ -0,0 +1,559 @@
...
+    return {
+      'value':       i.value,
+      'displayText': i.name,
+      'displayHtml': this._buildColorItemDisplayHtml(i),
+      'selected':    values.includes(i.value)
+    }
+  }, this));
+}
+
+// TODO
```
jcredding: @kellyredding - not needed anymore?

jcredding (approved):

PR #264, opened 2018-03-07
Allow appending popups (dropdowns/modals/tooltips) to a specific elem

jcredding:
This updates the dropdown, modal, and tooltip components to allow
appending their popups to a specific element. They already had
logic to append their popups to a different ancestor element but
this updates to allow them to append to any element on the page.
The main use for this is to allow mixing romo CSS and javascript
logic with legacy CSS and javascript. The popups can be appended
to a popup container elem that has the `romo` class instead of
having to add the `romo` class to the body elem which affects the
styles for the entire page.

This also updates the picker, select, and color select components
to allow using this option as well. They look for a component
specific data attr and write the dropdown append to data attr to
the dropdown elem they build. This allows their popups to be
appended to a special container the same as a regular dropdown,
modal, or tooltip.

### DIFF
--- a/assets/js/romo/color_select.js
+++ b/assets/js/romo/color_select.js
@@ -318,6 +318,20 @@ RomoColorSelect.prototype._buildOptionListDropdownElem = function() {
       Romo.data(this.elem, 'romo-color-select-dropdown-height')
     );
   }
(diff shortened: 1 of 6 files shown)


kellyredding (approved):
@jcredding changes look good.  This is a better option than the previous closest one.  Thanks. 👎

PR #265, opened 2018-03-09
rework color select item options data format

kellyredding:
This switches to using a list of tuples instead of a key-value
object.  This is less verbose and more concise to read/prepare.
It overall plays better when integrating in apps.  Plus the
overhead of the key names in the object format just add noise.

This also tweaks the the empty option building logic to use the
display text as the display html os that empty options actually
show a text value in the dropdown if one is given.  I noticed this
when trying to use this empty option logic in an app.

@jcredding ready for review.


jcredding (approved):

PR #266, opened 2018-03-21
Fix "ddd" `RomoDate.format` for the 11th, 12th, and 13th of the month

jcredding:
This updates the `RomoData.format` function to better handle the
"ddd" format string when the day is the 11th, 12th, or 13th of the
month. They are currently converting these days to the 11st, 12nd,
and 13rd. This fixes the issue by expliciting checking if the
day is 11, 12, or 13 and if so not applying the normal suffix
rules when the lasy day number is 1, 2, or 3.


kellyredding (approved):
@jcredding thanks for cleaning my mess up here 👎

